PR #186, opened 2024-12-07
fix: add missing period

nielsvoss:


nielsvoss:
Not sure if periods belong after inline code, it just seemed to be the case throughout the rest of the documentation.

david-christiansen (approved):

david-christiansen:
I think they do, when that code comes at the end of a sentence. It can make sense to avoid them in plain-text communication, but anything that allows the code literals to be demarcated (HTML, Markdown, TeX, etc) doesn't have the same risk of confusion.

Thanks!

PR #201, opened 2024-12-09
chore: ci: factor vale check out into separate workflow

david-christiansen:
This is a step towards getting deployments to work for external PRs


david-christiansen:
Superseded by #202

PR #229, opened 2025-01-06
universes: clarify wording that may sounds like universs are cumutative

nomeata:
The old phrasing sounded wrong to me. The new one is a bit jolty, so feel free to phrase differently.


david-christiansen:
I think it's great as-is. Thanks!

PR #234, opened 2025-01-08
feat: more on structural recursion

nomeata:
This PR extends the section on structural recursion. In particular it attempts to explain the rules first without going into the underlying construction, confining that existing discussion to a subsection.

The section on mutual recursion is quite terse, but better than nothing and a start.


david-christiansen:
Whoa! I spent time today writing the same section. I'll hold off on further work on it until you mark this as ready, then I'll merge my examples and explanations in where they make sense.

nomeata:
Oh sorry, that's an unfortunate misunderstanding!

PR #238, opened 2025-01-09
feat: fixed-precision integers

### LINKED ISSUE #105: Fixed-precision integers
**What question should the reference manual answer?**

How are the fixed-precision integer types represented in the kernel and at run time? What are their APIs?

david-christiansen:
Closes #105

### DIFF
--- a/Manual/BasicTypes/UInt.lean
+++ b/Manual/BasicTypes/UInt.lean
@@ -1,0 +1,592 @@
+/-
+Copyright (c) 2025 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: David Thrane Christiansen
+-/
+
+import VersoManual
+
+import Manual.Meta
+import Manual.BasicTypes.UInt.Comparisons
+import Manual.BasicTypes.UInt.Arith
+
+open Manual.FFIDocType
+
+open Verso.Genre Manual
+
+#doc (Manual) "Fixed-Precision Integer Types" =>
+%%%
+tag := "fixed-ints"
+%%%
+
+Lean's standard library includes the usual assortment of fixed-width integer types.
+From the perspective of formalization and proofs, these types are wrappers around bitvectors of the appropriate size; the wrappers ensure that the correct implementations of e.g. arithmetic operations are applied.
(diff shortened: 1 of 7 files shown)


Manual/BasicTypes/UInt.lean
```diff
@@ -0,0 +1,592 @@
...
+{docstring Int64}
+
+# Run-Time Representation
+
+In compiled code, fixed-width integer types that fit in one less bit than the platform's pointer size are represented unboxed, without additional allocations or indirections.
+This always includes {lean}`Int8`, {lean}`UInt8`, {lean}`Int16`, and {lean}`UInt16`.
+On 64-bit architectures, {lean}`Int32` and {lean}`UInt32` are also unboxed.
+
+Fixed-width integer types that take at least as many bits as the platform's pointer type are represented the same way as {name}`Nat`: if they are sufficiently small, then they are represented unboxed; if they are larger, then they are represented by a heap-allocated number value.
+{lean}`Int64`, {lean}`UInt64`, {lean}`ISize`, and {lean}`USize` are always represented this way, as are {lean}`Int32` and {lean}`UInt32` on 32-bit architectures.
```
eyelash, line 64: This paragraph seems to imply that on a 64 bit platform, `Int64`, `UInt64`, `ISize`, and `USize` are represented the same way as `Nat`. [The FFI documentation](https://lean-lang.org/lean4/doc/dev/ffi.html) however, says that they should be represented as `uint64_t` and `size_t` which would be different from `Nat` which is represented as `lean_object *`. Indeed the following function:

```lean
def foo (x : UInt64) : UInt64 := x + 1
```

gets compiled to the following C code:

```c
LEAN_EXPORT uint64_t l_foo(uint64_t x_1) {
_start:
{
uint64_t x_2; uint64_t x_3; 
x_2 = 1;
x_3 = lean_uint64_add(x_1, x_2);
return x_3;
}
}
```

and [`lean_uint64_add`](https://github.com/leanprover/lean4/blob/8791a9ce069d6dc87f7cccc4387545b1110c89bd/src/include/lean/lean.h#L1786) is just a simple addition of two `uint64_t`s so it seems to me these are not represented the same way as `Nat`.

PR #239, opened 2025-01-09
fix: minor typo/typography in Simp.lean

benjaminfjones:
This change adds a missing word and corrects one minor typographic error.


david-christiansen:
Thank you very much!

PR #241, opened 2025-01-10
feat: search box

david-christiansen:
Adds a search box that uses Verso's internal metadata for a live index.


david-christiansen:
Based on work in https://github.com/leanprover/verso/pull/267

ashandoak:
@jakobvase the live search results seem to be vertically misaligned. Looks like the term (on the left) is top justified, while the category (on the right) is bottom justified within the flex container.

Screenshot 2025-01-21 at 7 24 38 PM

jakobvase:
> @jakobvase the live search results seem to be vertically misaligned. Looks like the term (on the left) is top justified, while the category (on the right) is bottom justified within the flex container.
> Screenshot 2025-01-21 at 7 24 38 PM

Yeah that's how I wanted it to work. If you view it on mobile or a smaller screen, or if the term is long, the category is displayed below the term.

But you're right that on your screenshot it looks off. I'll see if I can improve that.

jakobvase:
Did a bit of testing. It looks good and works on mobile, but I should disable the spelling autocorrection. And I’m curious what you think of the search results?

david-christiansen:
The results seem basically fine here. I think there's one problem that sticks out: the results seem incomplete. Searching `Array` gets me many array operations, but not all of them. Is there a limit? Can we increase it, or at least indicate that results are truncated and give a way to ask for more?


I was surprised by a few things in the ranking, but they didn't scream "BUG" to me:
 * I would usually expect a fuzzy-search algorithm to prioritize bits of string that are typed contiguously over ones that don't. But when searching `Ind x`, `Fin.reduceXor` is prioritized over `Lean.Syntax.getKind` even though `ind` are not contiguous in the former.
 * It would be nice if we could down-prioritize parts of a string when ranking matches. For example, weighting matches in the actual name over matches in the namespace. I think that'd make the ordering more useful.
 * Searching for `!` also returns results for `1`. This makes a search for operations that might panic (e.g. `Option.get!` not so useful, because section numbers participate in searching and crowd the results. This is esp. bad with the limited number of results.

I think I need to add syntax tokens to the index, so that searching for e.g. `if` or `⟨` will make that come up. I think that'll be a v2 thing.

It seems to work well on my phone (thanks!), but the interaction with scrolling feels very "desktoppy". Another thing we could look at in a second version would be making the box into a modal on mobile screens.

jakobvase:
Yeah I set a limit of 20 to begin with. We can increase it, but if the list gets too long users get lost in the list and performance decreases.

I also agree with everything you write about the results. I think improving the results is one of the top goals for a v2.

I'll increase the limit to 30 and disable spelling corrections tonight, then I think we're set for release?

jakobvase:
I've increased the limit to 30, removed autocorrect on ios, and added the "showing n/m results".

I see the "!" as "1" too, but only some of the time. I haven't had the time to dig into it - I can see if I have time later this week, but I don't see it as a blocker. I agree it has to be fixed soon.

david-christiansen:
This is much nicer! Thank you!

jakobvase:
I've fixed the 1/! issue. Please remember to change the toc's z-index to 9, so it doesn't look like this: 
Screenshot 2025-01-29 at 17 07 04
Otherwise we're good for release from my side!

david-christiansen:
OK, I think it's finally good to go. Thanks again! 🚀

PR #250, opened 2025-01-14
feat: well-founded recursion

### LINKED ISSUE #57: Describe well-founded recursion
The section should include:

 * How a well-founded relation is chosen
 * Accessibility
 * `termination_by` and `termination_by?`
 * Specify the properties of well-founded recursion that are relevant for practical use of Lean

nomeata:
Closes #57


Manual/Language/RecursiveDefs/WF.lean
```diff
@@ -0,0 +1,216 @@
+/-
+Copyright (c) 2024 Lean FRO LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Author: David Thrane Christiansen
```
david-christiansen: Surely this should read
```suggestion
Author: Joachim Breitner
```

Manual/Language/RecursiveDefs/WF.lean
```diff
@@ -0,0 +1,216 @@
+/-
+Copyright (c) 2024 Lean FRO LLC. All rights reserved.
```
david-christiansen:
```suggestion
Copyright (c) 2025 Lean FRO LLC. All rights reserved.
```

david-christiansen:
I'll wait until it's un-drafted to read the text. Thanks!

nomeata:
Ok, I think that’s a decent first iteration.

It is much less detailed on the internal construction than the structural recursion section, but think that’s ok, and the construction doesn’t have to be covered that prominently.

Please give it a read when you get to it, and either let me know what's missing (and I can write some more), or take it over get it into final shape, what works better for you.

It maybe be that I am not using Verso’s full potential when it comes to showing proofs states, I’ll watch your changes there and learn.

Manual/Language/RecursiveDefs/WF.lean
```diff
@@ -14,16 +14,27 @@ open Verso.Genre
...
+All definitions are treated identically by the Lean compiler.
+In Lean's logic, definitions that use well-founded recursion do not necessarily reduce {tech key:="definitional equality"}[definitionally].
+The reductions do hold as propositional equalities, however, and Lean automatically proves them.
+This does not typically make it more difficult to prove properties of definitions that use well-founded recursion, because the propositional reductions can be used to reason about the behavior of the function.
+It does mean, however, that using these functions in types typically does not work well.
+When the reduction behavior does hold definitionally, it is often much slower than structurally recursive definitions in the kernel, which must unfold the termination proof along with the definition.
+When possible, recursive function that are intended for use in types or in other situations where definitional equality is important should be defined with structural recursion.
```
nomeata: I’m inclined to phrase this more strongly: Definitional equality for wf recursion is an artifact that should not be relied upon.

```suggestion
All definitions are treated identically by the Lean compiler.
In Lean's logic, definitions that use well-founded recursion typically do not reduce {tech key:="definitional equality"}[definitionally].
The reductions do hold as propositional equalities, however, and Lean automatically proves them.
This does not typically make it more difficult to prove properties of definitions that use well-founded recursion, because the propositional reductions can be used to reason about the behavior of the function.
It does mean, however, that using these functions in types typically does not work well.
Even when the reduction behavior happens to hold definitionally, it is often much slower than structurally recursive definitions in the kernel, which must unfold the termination proof along with the definition.
When possible, recursive function that are intended for use in types or in other situations where definitional equality is important should be defined with structural recursion.
```

Manual/Language/RecursiveDefs/WF.lean
```diff
@@ -14,16 +14,27 @@ open Verso.Genre
...
 
-To explicitly use well-founded recursion recursion, a function or theorem definition can be annotated with a {keywordOf Lean.Parser.Command.declaration}`termination_by` clause that specifies the {deftech}_termination argument_.
+To explicitly use well-founded recursion recursion, a function or theorem definition can be annotated with a {keywordOf Lean.Parser.Command.declaration}`termination_by` clause that specifies the {deftech}_measure_ by which the function terminates.
+The measure should be a term that decreases at each recursive call; it may be one of the function's parameters or a product of the parameters, but it may also be any other term.
```
nomeata:
```suggestion
The measure should be a term that decreases at each recursive call; it may be one of the function's parameters or a tuple of the parameters, but it may also be any other term.
```
Maybe avoiding “product” near so many `Nat`, else the reader is confused why `n * m` is a measure worth pointing out here.

PR #251, opened 2025-01-15
chore: Use on workflow_run

nomeata:
instead of reacting on label changes. This copies the setup found in
https://github.com/leanprover/reference-manual/blob/main/.github/workflows/pr-deploy.yml


nomeata:
This is tedious to test outside production. Merge and iterate, or set up a test repo with everything, debug there, and only then merge here?

david-christiansen:
Please feel free to merge and iterate! Thank you!

PR #252, opened 2025-01-16
chore: dummy change to check CI setup

nomeata:


nomeata:
I couldn’t get the nicer neltify-deploy-related notifications to work, so am using a comment-posting action, as can be seen above. Maybe good enough? Or we can revert, go back to the labeling approach, and investiate why the removing and adding the label didn’t work (maybe adding sleep in between helps).

PR #253, opened 2025-01-16
feat: partial_fixpoint

nomeata:
The manual section to go with https://github.com/leanprover/lean4/pull/6355.


Manual/Meta/Monotonicity.lean
```diff
@@ -0,0 +1,75 @@
...
+            throwError "Unexpecte conclusion of {name}"
+          let f := concl.appArg!
+          unless f.isLambda do
+            throwError "Unexpecte conclusion of {name}"
+          lambdaBoundedTelescope f 1 fun _ _call => do
+            -- Could not get this to work:
+            -- let _stx ← Lean.PrettyPrinter.delab call
+            -- `(Inline.code $(quote stx))
+            `(Inline.text "TODO")
+            -- pure (some stx)
```
nomeata: I dug my toes into the `Manual.Meta` realm, because I’d like to auto-generate a list of registered monotonicity lemmas. This worked fine, and I found how to pretty print the name of the lemma. But I’d also include some expressions that I want to calculate from the type. The above didn’t work, probably because of one level of quotation too little.

I guess I could pretty-print to a string and include that…

But can I also “delaborate” an Expr to something that I can put in `Inline.code` so that it’s highlighted and hovers work?

nomeata: ~~I also tried to use a table instead of a `ul`, but my attempt at https://github.com/leanprover/reference-manual/pull/253/commits/56b0219896008de8a350b4ff6d86681e3ddb3e9a also ran into my ignorance of how verso works. Maybe we can have a quick look together next week, during tea time or else.~~

nomeata:
Direct link to new section: https://draft-pr-253--lean-reference-manual-review.netlify.app/Recursive-Definitions/Partial-Fixpoint-Recursion/#partial-fixpoint

nomeata:
Ok, marking this as ready for review for now, and will try to keep my hands away from it.

I’ll merge the feature to Lean4 `master` next, but presumably this PR should not enter `main` until the feature is in a pre-release? So there is no rush with getting this into shape.

I left some `todo` notes where my Verso-skills failed me, but none are critical, they are all just nice-to haves.

PR #268, opened 2025-01-30
fix: remove a typo in WF.lean

sgraf812:
Just a redundant "recursion". Although the fact that the term "induction-recursion" exists makes me wonder if this isn't actually a typo :)


david-christiansen:
Thank you!

PR #275, opened 2025-02-03
feat: improvements to Quotient chapter suggested by Kyle

david-christiansen:


Manual/Quotients.lean
```diff
@@ -27,6 +27,9 @@ In particular, all functions from a quotient type must prove that they respect t
 
 {docstring Quotient}
 
+A proof that two elements of the underlying type are related by the equivalence relation is sufficient to prove that they are equal in the {name}`Quotient`.
+However, {tech}[definitional equality] is unaffected by the use of {lean}`Quotient`: two elements in the quotient are definitionally equal if and only if they are equal in the underlying type.
```
kmill:
```suggestion
However, {tech}[definitional equality] is unaffected by the use of {lean}`Quotient`: two elements in the quotient are definitionally equal if and only if they are definitionally equal in the underlying type.
```

kmill (approved):

PR #291, opened 2025-02-07
fix: correct the description of integer representation

### LINKED ISSUE #286: Runtime representation of fixed-width types is described incorrectly
**Describe the bug**

From [here](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#The-Lean-Language-Reference--Basic-Types--Fixed-Precision-Integer-Types--Run-Time-Representation):

> In compiled code, fixed-width integer types that fit in one less bit than the platform's pointer size are represented unboxed, without additional allocations or indirections. This always includes [Int8](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#Int8___mk), [UInt8](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#UInt8___mk), [Int16](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#Int16___mk), and [UInt16](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#UInt16___mk). On 64-bit architectures, [Int32](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#Int32___mk) and [UInt32](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#UInt32___mk) are also unboxed.
> 
> Fixed-width integer types that take at least as many bits as the platform's pointer type are represented the same way as [Nat](https://lean-lang.org/doc/reference/latest/Basic-Types/Natural-Numbers/#Nat___zero): if they are sufficiently small, then they are represented unboxed; if they are larger, then they are represented by a heap-allocated number value. [Int64](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#Int64___mk), [UInt64](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#UInt64___mk), [ISize](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#ISize___mk), and [USize](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#USize___mk) are always represented this way, as are [Int32](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#Int32___mk) and [UInt32](https://lean-lang.org/doc/reference/latest/Basic-Types/Fixed-Precision-Integer-Types/#UInt32___mk) on 32-bit architectures.

In what way is the content incorrect?

The text describes the representation of fixed-width integers if they appear in a place where an object is expected (for example, in an `Array UInt64`). However, in most situations (like in structures, or in function arguments and return types, or in intermediate results during function evaluation), a type is represented exactly by the corresponding fixed-width C type as described in [ffi.md](https://github.com/leanprover/lean4/blob/master/doc/dev/ffi.md).

**Demonstration**

Here is a Lean program:

```lean
structure A where
  i : UInt64

structure B where
  i : UInt64
  j : Nat

def f (arr : Array UInt64) (h : 0 < arr.size) : UInt64 :=
  arr[0]

def g (a : A) : UInt64 :=
  a.i

def h (b : B) : UInt64 :=
  b.i

def i (arr : Array UInt64) (harr : 0 < arr.size) (a : A) (b : B) (bare : UInt64) : UInt64 :=
  f arr harr + g a + h b + bare

def main : IO Unit :=
  return ()
```

We can compile this using `lake build` and look at the generated C code. Let's look at the code for the function `i`:

```c
LEAN_EXPORT uint64_t l_i(lean_object* x_1, lean_object* x_2, uint64_t x_3, lean_object* x_4, uint64_t x_5) {
_start:
{
lean_object* x_6; lean_object* x_7; uint64_t x_8; uint64_t x_9; uint64_t x_10; uint64_t x_11; uint64_t x_12; 
x_6 = lean_unsigned_to_nat(0u);
x_7 = lean_array_fget(x_1, x_6);
x_8 = lean_unbox_uint64(x_7);
lean_dec(x_7);
x_9 = lean_uint64_add(x_8, x_3);
x_10 = lean_ctor_get_uint64(x_4, sizeof(void*)*1);
x_11 = lean_uint64_add(x_9, x_10);
x_12 = lean_uint64_add(x_11, x_5);
return x_12;
}
}
```
From this we learn:

- Both the structure `A` (argument `x_3`) and the type `UInt64` (argument `x_5`) are represented as `uint64_t` when passed to the function
- We get a `uint64_t` out of the structure `B` directly using `lean_ctor_get_uint64` (cf. the `ffi.md`)
- When we extract a `UInt64` from an `Array UInt64`, we get `lean_object` which we have to turn into a `uint64_t` using `lean_unbox_uint64`. This is what is described in the reference manual above.

(NB: `lean_uint64_add` is implemented as `return a1+a2;`)

Looking at `f`, `g`, and `h`:
```c
LEAN_EXPORT lean_object* l_f(lean_object* x_1, lean_object* x_2) {
_start:
{
lean_object* x_3; lean_object* x_4; 
x_3 = lean_unsigned_to_nat(0u);
x_4 = lean_array_fget(x_1, x_3);
return x_4;
}
}

LEAN_EXPORT uint64_t l_g(uint64_t x_1) {
_start:
{
return x_1;
}
}

LEAN_EXPORT uint64_t l_h(lean_object* x_1) {
_start:
{
uint64_t x_2; 
x_2 = lean_ctor_get_uint64(x_1, sizeof(void*)*1);
return x_2;
}
}
```
The main interesting thing to note here is that `f` returns an object, while `g` and `h` return `uint64_t`s, even though all of them return `UInt64` in the Lean world. So there is no "eager unboxing" in the case of `f`.

david-christiansen:
Closes #286

### DIFF
--- a/Manual/BasicTypes/UInt.lean
+++ b/Manual/BasicTypes/UInt.lean
@@ -64,9 +64,9 @@ On 32-bit architectures, {lean}`Int32` and {lean}`UInt32` are boxed, which means
 
 Even though some fixed-with integer types require boxing in general, the compiler is able to represent them without boxing in code paths that use only a specific fixed-width type rather than being polymorphic, potentially after a specialization pass.
 This applies in most practical situations where these types are used: their values are represented using the corresponding unsigned fixed-width C type when a constructor parameter, function parameter, function return value, or intermediate result is known to be a fixed-width integer type.
-The Lean run-time system includes primitives for storing fixed-width integers in datatype constructors, and the primitive operations are defined on the corresponding C types, so boxing tends to happen at the “edges” of integer calculations rather than for each intermediate result.
+The Lean run-time system includes primitives for storing fixed-width integers in constructors of {tech}[inductive types], and the primitive operations are defined on the corresponding C types, so boxing tends to happen at the “edges” of integer calculations rather than for each intermediate result.
 In contexts where other types might occur, such as the contents of polymorphic containers like {name}`Array`, these types are boxed, even if an array is statically known to contain only a single fixed-width integer type.{margin}[The monomorphic array type {lean}`ByteArray` avoids boxing for arrays of {lean}`UInt8`.]
-Lean does not specialize the representation of datatypes or arrays.
+Lean does not specialize the representation of inductive types or arrays.
 Inspecting a function's type in Lean is not sufficient to determine how fixed-width integer values will be represented, because boxed values are not eagerly unboxed—a function that projects an {name}`Int64` from an array returns a boxed integer value.
 
 # Syntax
--- a/Manual/BasicTypes/UInt.lean
+++ b/Manual/BasicTypes/UInt.lean
@@ -58,15 +58,16 @@ Signed integers wrap the corresponding unsigned integers, and use a twos-complem
 
 In compiled code, fixed-width integer types that fit in one less bit than the platform's pointer size are always represented unboxed, without additional allocations or indirections.
 This always includes {lean}`Int8`, {lean}`UInt8`, {lean}`Int16`, and {lean}`UInt16`.
-On 64-bit architectures, {lean}`Int32` and {lean}`UInt32` are also always unboxed.
+On 64-bit architectures, {lean}`Int32` and {lean}`UInt32` are also unboxed.
+On 32-bit architectures, {lean}`Int32` and {lean}`UInt32` are boxed, which means they may be represented by a pointer to an object on the heap.
+{lean}`ISize`, {lean}`USize`, {lean}`Int64` and {lean}`UInt64` are boxed on all architectures.
 
-In most situations, all of the fixed-width integer types are unboxed.
-Their values are represented using the corresponding unsigned fixed-width C type when a constructor parameter, function parameter, function return value, or intermediate result is known to be a fixed-width integer type.
-In contexts where other types might occur, such as the contents of polymorphic containers like {name}`Array`, these types are boxed.
+Even though some fixed-with integer types require boxing in general, the compiler is able to represent them without boxing in code paths that use only a specific fixed-width type rather than being polymorphic, potentially after a specialization pass.
+This applies in most practical situations where these types are used: their values are represented using the corresponding unsigned fixed-width C type when a constructor parameter, function parameter, function return value, or intermediate result is known to be a fixed-width integer type.
+The Lean run-time system includes primitives for storing fixed-width integers in datatype constructors, and the primitive operations are defined on the corresponding C types, so boxing tends to happen at the “edges” of integer calculations rather than for each intermediate result.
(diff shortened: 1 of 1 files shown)


david-christiansen:
@TwoFX: Thanks for the great issue!

One thing I'm not totally happy about with this version of the text is that it uses the term "unboxed" in a somewhat ambiguous manner. I'm trying to balance conceptual clarity against overwhelming readers with pedantic details, but I'm not sure this strikes the best balance - what do you think?

In particular, "unboxed" is meaning two different things:
 1. Representing small values inside a pointer, rather than on the heap, in a `lean_object*`
 2. Using a completely different, definitely non-pointer type like `uint64_t` based on flow-sensitive properties of monomorphic contexts

In both cases, there's neither heap allocation nor indirection, so in that sense they're the same.

TwoFX:
I had another look at `lean.h` and I'm no longer sure I believe the text as written. Here are the boxing and unboxing functions:
```c
static inline lean_object * lean_box(size_t n) { return (lean_object*)(((size_t)(n) << 1) | 1); }
static inline size_t lean_unbox(lean_object * o) { return (size_t)(o) >> 1; }

static inline lean_obj_res lean_box_uint32(uint32_t v) {
    if (sizeof(void*) == 4) {
        /* 32-bit implementation */
        lean_obj_res r = lean_alloc_ctor(0, 0, sizeof(uint32_t));
        lean_ctor_set_uint32(r, 0, v);
        return r;
    } else {
        /* 64-bit implementation */
        return lean_box(v);
    }
}

static inline unsigned lean_unbox_uint32(b_lean_obj_arg o) {
    if (sizeof(void*) == 4) {
        /* 32-bit implementation */
        return lean_ctor_get_uint32(o, 0);
    } else {
        /* 64-bit implementation */
        return lean_unbox(o);
    }
}

static inline lean_obj_res lean_box_uint64(uint64_t v) {
    lean_obj_res r = lean_alloc_ctor(0, 0, sizeof(uint64_t));
    lean_ctor_set_uint64(r, 0, v);
    return r;
}

static inline uint64_t lean_unbox_uint64(b_lean_obj_arg o) {
    return lean_ctor_get_uint64(o, 0);
}
```

`lean_box` and `lean_unbox` are generated for `Nat` as well as `UIntX` for `X <= 16`. For `UInt32` and `UInt64`, the compiler generates calls to the special functions.

So to me this looks like the statement "When boxed, fixed-width integer types that take at least as many bits as the platform's pointer type are represented the same way as `Nat`" is not correct for 64-bit integers on 64-bit systems and for 32/64-bit integers on 32-bit systems, since these always have an indirection, not just for small values.

david-christiansen:
Thank you for the thorough investigation and correction of misunderstandings - I think this is convincing.

What do you think of this update?

TwoFX:
I really like this latest version!

david-christiansen:
Thanks again for the very useful help!

PR #309, opened 2025-02-16
feat: improve legibility

jakobvase:
This aims to make the reference manual appear more modern.

> use sans-serif for most text: Serif typefonts are not meant for display on the web, except in specific
cases.

> add roboto to the list of fonts: Helvetica and Segoe are not available on android devices.


david-christiansen:
The choice of the Source family of fonts here was intentional, because they have compatible metrics. In particular, Source Code and Source Sans have the same x-height. Without this property, text that mentions identifiers looks janky and mismatched. I would think we should use Source Sans as the first choice in the list here.

ashandoak:
@david-christiansen you're referencing the change to `--verso-text-font-family` and recommending that the first item on this list should be Source Sans?

jakobvase:
I've added Source Sans 3 to the repo (copied from https://github.com/adobe-fonts/source-sans). I think it's the same place the other fonts were taken from, and the licenses are the same.

I've also added Source Sans 3 to the list of fonts for `--verso-text-font-family`.

Finally, I added quotes around Source Code Pro, because the spec recommends adding quotes to fonts with whitespace. https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#propdef-font-family.

david-christiansen:
@ashandoak Yes, that's right!

@jakobvase This looks great, thanks!

Issue #318, opened 2025-02-22
Add a header to the reference manual?


jakobvase:
Should the reference manual have a header? That would be a place to put the menu icons on mobile, the quick-jump box, and the title of the book - also relevant for other Verso projects.

Issue #317, opened 2025-02-22
Improvements to the table of contents


jakobvase:
The table of contents has a few issues that make it slightly confusing to use.

- Opening and closing of the 'fold out' lists is inconsistent. When navigating, it should keep the parent of the current leaf of the tree open and close the rest, so it's easy to see where you are.
- You should be able to see all the parents when scrolling in the toc, so you always have the path to the things you're looking at.
- "The Lean Language Reference" is very verbose. Should maybe be moved to a header?
- The cross and burger icons for the menu should be removed on desktop - the sidebar should always be open on desktop.
- It should be a bit more distinct what you're currently looking at. The dotted underline is very subtle.
- Figure out some other solution to the 'next page/previous page'. It looks a little too homemade, I think. And at least for the reference manual, I don't think the ability to navigate to a previous or a next page is very important.

david-christiansen:
For the collapsing sections, it should be a matter of deleting the JavaScript that remembers the state that the user left it in last time :⁠-⁠)

jakobvase:
Nice, that pointed me to why the sidebar animation was showing up when navigating on mobile. Thanks :)

david-christiansen:
These items are all completed now, right?

jakobvase:
Yes, or another solution was implemented.

PR #320, opened 2025-02-22
feat: professionalize the ui

### LINKED ISSUE #316: Make the index page less confusing
https://lean-lang.org/doc/reference/latest/the-index/#The-Lean-Language-Reference--Index has a list of terms from all over the site, but it's presented in a way that makes it hard to find what you're looking for.

Please make it a single list instead.

### LINKED ISSUE #319: Tighten up the presentation of definitions
If you look here: [https://lean-lang.org/doc/reference/latest/find/?domain=Verso.Genre.Manual.doc&name=String.toNat!](https://lean-lang.org/doc/reference/latest/find/?domain=Verso.Genre.Manual.doc&name=String.toNat!), you'll see the definition of `String.toNat!`. It takes up a lot of space for a header and two lines of description. I think we should make it smaller, so it becomes easier to find the definition you're looking for.

Something like this would be nice:

Image

That's with these css rules applied:

```css
.namedocs .text > :last-child,
.namedocs .text > :last-child * {
    margin-bottom: 0;
}

.namedocs .text > :first-child {
    margin-top: 0;
}
```

jakobvase:
This has a bunch of ideas related to issues #316 , #317 , #318 , and #319 . Closes #319. Closes #316.

It also points verso to a branch in my verso-repo, which is why it's in draft and shouldn't be merged - also because there are a lot of changes and maybe some of them shouldn't get in.

It builds on top of the sans-serif pull request, so this is the combined changes.

Items addressed:
- Add a fixed header
- Reduce font size on desktop
- Increase line height for body content
- Make margins on all sides of definition boxes the same size.
- Decrease the font size of the definition header text
- Reduce index header font size
- Make the index header not stick on mobile
- Make the index always single-column
- Disable toc hide/show on desktop

I'll add more tomorrow.


static/theme.css
```diff
@@ -1,7 +1,7 @@
 :root {
-    --verso-structure-font-family: "Helvetica Neue","Segoe UI",Arial,sans-serif;
-    --verso-text-font-family: Source Serif, Georgia,Times,"Times New Roman",serif;
-    --verso-code-font-family: Source Code Pro, monospace;
+    --verso-structure-font-family: "Helvetica Neue","Segoe UI","Roboto",Arial,sans-serif;
```
david-christiansen: Shouldn't this also include Source Sans 3 as the first entry?

jakobvase: Hmm. It wasn't before, so I'm not sure. I'll compare them in the weekend.

jakobvase:
I've rebased and fixed the verso pointer.

david-christiansen:
Rebased on `main` to fix the internal LSP API incompatibility that was blocking CI

static/theme.css
david-christiansen: It wasn't before because it didn't seem useful to distribute a font just for headings that matched the metrics with the code font. But now that the body text is sans, I think the header font should be the same font.

david-christiansen:
All right, looks like we're good!

PR #336, opened 2025-03-08
fix: make distance between list element title and sublist the same

jakobvase:
This has annoyed me for a while. It seems something has changed in the rendering of the contents on the chapter-pages, so they don't show the sub-lists anymore. But the index still does. This changes the index from looking like this:

Screenshot 2025-03-08 at 08 31 19

To looking like this:

Screenshot 2025-03-08 at 08 31 28

### DIFF
--- a/static/theme.css
+++ b/static/theme.css
@@ -55,6 +55,9 @@ main section li:not(:first-child) {
 main section li:not(:last-child) {
   margin-bottom: 0.5rem;
 }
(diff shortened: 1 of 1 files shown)


david-christiansen:
That's right, the super-detailed overview of subsections on those pages was made redundant by the ToC improvements, and it mostly was overwhelming, so they're gone.

This looks good in your screenshot.

PR #348, opened 2025-03-16
fix: couple the term and domain in the quick jump bar tighter on mobile

jakobvase:
Screen Shot 2025-03-16 at 13 51 11

I'd like a second opinion before we merge this. I think it's better than what we have currently for mobile, but I think @ashandoak should take a look too.


ashandoak:
Oh ya, much better. Great improvement @jakobvase!

david-christiansen:
Nice!

PR #350, opened 2025-03-17
feat: add release notes

nomeata:
This is a MVP for including the release notes in the reference manual, see
https://pr-350--lean-reference-manual-review.netlify.app/releases/v4.17.0/
for a preview.

It uses Verso’s markdown parser so that one does not have to manually edit the text for verso-compatibility (our commit messages are messy).

It uses one file and page per release, for easy linking and that it doesn’t grow too large, and (in particular) that you can scroll or Ctrl-F without losing track of which version this relates to.


david-christiansen:
If you're parsing the Markdown anyway, then perhaps it's a better idea to not require a Lean file for each version?

Then the code could be something like:
```
# Release Notes

{includeMarkdown "release-notes/4.18.md"}

{includeMarkdown "release-notes/4.17.md"}
```

or even

```
# Release Notes

{includeMarkdownDir "release-notes"}
```

nomeata:
I thought about it, but I’m worried about confusion because Metacode cannot tell lake about external dependencies, can it?

nomeata:
This may be good enough for a first iteration of including the release notes in the manual.

One reason why I opted for the ````` ```markdown ````` block is that the release notes are messy, and verso’s markdown-like language is more strict, and falls over the `_` in bv_decide, and other non-escaped fragments of code.

david-christiansen:
Given that the code in the Markdown block isn't expected to be rendered with highlights, they can processed at run time rather than elaboration time. That avoids issues with Lake not knowing about non-Lean files.

That said, it's a minor usability improvement.

Looking at the code, my only real concern here is that it's not generating actual headers from the Markdown. This means that the internal navigation will probably not work. The function you're calling is really written for hovers, rather than for proper documents, which don't have the same kind of nesting structure as proper documents, but it should be reasonably straightforward to write a little wrapper around it that generates a `Part` from Markdown headers. Happy to help if you get stuck.

If there's just one set of release notes, lack of navigation isn't a big deal, but it will matter when we get more.

nomeata:
> Given that the code in the Markdown block isn't expected to be rendered with highlights, they can processed at run time rather than elaboration time. That avoids issues with Lake not knowing about non-Lean files.

Ah, runtime! Good point.

One benefit of having it here is that one easily manually polish up bits of it, e.g. adjust at least the highlights section to properly refer to the rest of the manual.

> Looking at the code, my only real concern here is that it's not generating actual headers from the Markdown. This means that the internal navigation will probably not work. The function you're calling is really written for hovers, rather than for proper documents, which don't have the same kind of nesting structure as proper documents, but it should be reasonably straightforward to write a little wrapper around it that generates a `Part` from Markdown headers. Happy to help if you get stuck.

Yes, I noticed.

 
> If there's just one set of release notes, lack of navigation isn't a big deal, but it will matter when we get more.

My vision here is to have one file and page per release, for easy linking and that it doesn’t grow too large, and (in particular) that you can scroll or Ctrl-F without losing track of which version this relates to

So I’d say better internal navigation within one version is nice-to-have, but not a blocker.

Manual/Releases.lean
```diff
@@ -0,0 +1,26 @@
...
+import Manual.Releases.«v4.17.0»
+
+open Manual
+open Verso.Genre
+
+#doc (Manual) "Release Notes" =>
+%%%
+tag := "release-notes"
+file := "releases"
+%%%
```
david-christiansen, line 41:
```suggestion
number := false
%%%
```
I think this is more like an appendix or an index than it is like content in the manual, and this would be more clear if it were outside the numbering system.

I also found it a bit hard to focus on the section titles when they were next to the section numbers - both are sequences of digits and dots. I think this would solve that too.

Manual/Releases.lean
```diff
@@ -0,0 +1,26 @@
...
+
```
david-christiansen, line 42: Can there be a paragraph of text here that explains what the section is? Right now, the page is a bit jarring. A few sentences ought to do the trick.

Manual/Releases/v4.15.0.lean
```diff
@@ -0,0 +1,665 @@
...
+open Manual
+open Verso.Genre
+
+
+#doc (Manual) "v4.15.0" =>
+%%%
+tag := "release-v4.15.0"
+file := "v4.15.0"
+%%%
+
```
david-christiansen, line 20: In the original context, these pages have dates associated with them. What about adding the release date here?

david-christiansen:
I think it's basically fine modulo section numbers, but I would like to revisit it later to add proper sub-headers for consistency with the rest of the manual. No need to let the perfect be the enemy of the good, though.

And what about calling the sections "Lean 4.17.0" rather than "v4.17.0"? I think this may be easier to figure out if someone stumbles onto the page.

Manual/Releases.lean
nomeata: Absolutely – just like https://rocq-prover.org/doc/V8.20.0/refman/appendix/history-and-changes/index.html.

nomeata:
> And what about calling the sections "Lean 4.17.0" rather than "v4.17.0"? I think this may be easier to figure out if someone stumbles onto the page.

Good idea, or maybe even “Lean Release 4.17.0”.

nomeata:
Updated.

I put the date into the title; the motivation is that this way the Release Notes section TOC gives a nice quick way to relate version number and dates.

We don't have the feature (known from LaTeX) where one can give a shorter title to be used in ToCs, right? (Then I'd omit the “Lean” from “Lean 4.17.0 (2025-03--3)” in the TOCs where it’s nested under “Release Notes”)

david-christiansen:
The short ToC title feature would be nice to have, and pretty easy to implement.

First, add a new `Option String`-valued field [here](https://github.com/leanprover/verso/blob/36314ba11b9e12599e36c90d5b1ab5fa55fedac0/src/verso-manual/VersoManual/Basic.lean#L204-L216).

Then, set up the code that generates ToCs to check that field. Here's [one place](https://github.com/leanprover/verso/blob/36314ba11b9e12599e36c90d5b1ab5fa55fedac0/src/verso-manual/VersoManual.lean#L280-L286), which may be enough.

Manual/Releases.lean
```diff
@@ -0,0 +1,35 @@
...
+%%%
+tag := "release-notes"
+file := "releases"
+number := false
+%%%
+
+This section provides release notes about recent versions of Lean and is presented in reverse
+chronological order. When updating your copy of Lean to a new version, it is strongly recommended
+that you read the corresponding release notes. They may contain advice that will help you understand
+the differences with the previous version and upgrade your projects.
```
david-christiansen:
```suggestion
This section provides release notes about recent versions of Lean. When updating to a new version, please
read the corresponding release notes. They may contain advice that will help you understand
the differences with the previous version and upgrade your projects.
```

NItpicky rationales:
 * Readers can see that it's reverse order, I'm not sure it contributes much.
 * The user's copy of Lean doesn't get updated, really - the new version ends up next to it. This is admittedly nitpicky :)
 * "It is strongly recommended that" - this is a use of the passive voice that I think makes it harder to follow. I rephrased it from an existential claim that a recommendation exists into just being the recommendation :-)

Manual/Releases.lean
```diff
@@ -0,0 +1,35 @@
...
+
+The work-in-progress release notes of a release candidate can be found [in the Lean Github
```
david-christiansen:
```suggestion
The work-in-progress release notes of a release candidate can be found [in the Lean GitHub
```

david-christiansen (approved):

Manual/Releases.lean
nomeata: I basically just copied https://rocq-prover.org/doc/V8.20.0/refman/changes.html :-)

PR #403, opened 2025-03-27
feat: partial and unsafe definitions

### LINKED ISSUE #59: `partial` and `unsafe`
Describe `partial` and `unsafe` definitions:

 * Interaction with the kernel and elaborator
 * What guarantees are there, and what aren't there?
 * How to bridge from unsafe to safe code?

david-christiansen:
Closes #59

### DIFF
--- a/Manual/RecursiveDefs.lean
+++ b/Manual/RecursiveDefs.lean
@@ -180,15 +180,110 @@ After the first step of elaboration, in which definitions are still recursive, a
 tag := "partial-unsafe"
 %%%
 
-:::planned 59
-This section will describe `partial` and `unsafe` definitions:
+
+While most Lean functions can be reasoned about in Lean's type theory as well as compiled and run, definitions marked {keyword}`partial` or {keyword}`unsafe` cannot be meaningfully reasoned about.
+From the perspective of the logic, {keyword}`partial` functions are opaque constants and {keyword}`unsafe` definitions do not exist.
+In exchange for the inability to use these functions for reasoning, there are far fewer requirements placed on them; this can make it possible to write programs that would be impractical or cost-prohibitive to prove anything about, while not giving up formal reasoning for the rest.
+In essence, the {keyword}`partial` subset of Lean is a traditional functional programming language that is nonetheless deeply integrated with the theorem proving features, and the {keyword}`unsafe` subset features the ability to break Lean's runtime invariants if needed, at the cost of less integration with Lean's theorem-proving features.
+Analogously, {keyword}`noncomputable` definitions may use features that don't make sense in programs, but are meaningful in the logic.
+
+## Partial Functions
+
+The {keyword}`partial` modifier may only be applied to function definitions.
+Partial functions are not required to demonstrate termination, and Lean does not attempt to do so.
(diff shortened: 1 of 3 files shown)


Manual/RecursiveDefs.lean
```diff
@@ -180,16 +180,111 @@ After the first step of elaboration, in which definitions are still recursive, a
...
+Analogously, {keyword}`noncomputable` definitions may use features that don't make sense in programs, but are meaningful in the logic.
 
- * Interaction with the kernel and elaborator
- * What guarantees are there, and what aren't there?
- * How to bridge from unsafe to safe code?
+## Partial Functions
 
+The {keyword}`partial` modifier may only be applied to function definitions.
+Partial functions are not required to demonstrate termination, and Lean does not attempt to do so.
+These functions are “partial” in the sense that they do not necessarily specify mapping from each element of the domain to an element of the codomain, because they might fail to terminate for some or all elements of the domain.
```
nomeata:
```suggestion
These functions are “partial” in the sense that they do not necessarily specify a mapping from each element of the domain to an element of the codomain, because they might fail to terminate for some or all elements of the domain.
```

Manual/RecursiveDefs.lean
```diff
@@ -180,16 +180,111 @@ After the first step of elaboration, in which definitions are still recursive, a
...
+They are elaborated into {tech}[pre-definitions] that contain explicit recursion, and type checked using the kernel; however, they are subsequently treated as opaque constants by the logic.
```
nomeata, line 198: Are they really type checked by the kernel? I'd assume that the kernel only sees the opaque definition with the inhabitant witnesses, but never the recursive rhs. (Didn't check, reading on the phone while waking up…)

Thinking of this line here:
https://github.com/leanprover/lean4/blob/060b2fe46fe728934744ee52271b92ab74cbd5b4/src/Lean/Elab/PreDefinition/Main.lean#L28

Manual/RecursiveDefs.lean
david-christiansen: We crossed messages on this one! Thanks though!

Manual/RecursiveDefs.lean
david-christiansen: I initially thought that they weren't, and tried to build an exploit based on an elaborator that cheated and got kernel errors, which led me to investigate :)

It's actually generating an unsafe recursive auxiliary function, which is kernel-checked in the way described here. That seemed to be an implementation detail rather than something users need to know, though.

Right after that line, `addAndCompilePartialRec` is called.
This generates an unsafe auxiliary recursive definition for each predef, then calls `addAndCompileUnsafe` on the auxiliary def, which calls `addDecl`, which bottoms out in `lean_elab_add_decl`. It calls this function:
https://github.com/leanprover/lean4/blob/060b2fe46fe728934744ee52271b92ab74cbd5b4/src/kernel/environment.cpp#L163-L178
which is where the recursive check is happening.

I'm pretty sure this is correct, if not a complete accounting of the internals.

david-christiansen: (and thanks!)

nomeata: Oh, I stand corrected! I saw unsafe in the function name and didn't look further. TIL!

david-christiansen: Thanks for your careful attention :)

PR #412, opened 2025-04-03
feat: release notes for v4.18.0

kim-em:


david-christiansen:
Thanks!

The Markdown had some missing nesting, which led to the page being rendered incorrectly, so I fixed it.

PR #447, opened 2025-05-23
chore: add highlights for v4.20 release notes

eyihluyc:
This PR adds highlights section to v4.20 release notes.


nomeata (approved):
Ha, there are even highlights that I missed :-) Thanks for that!

I’d remove `#8102`; this is not really a highlight, more a bug fix.

PR #462, opened 2025-06-05
chore: add #7428 to v4.21.0 changelog

Vierkantor:
Lean PR 7428 brings a huge speedup to `simp` so we should be sure to mention it!


david-christiansen:
Thanks!

PR #479, opened 2025-06-24
feat: allow examples to be open by default

jrr6:
This PR adds a field to the `example` block configuration allowing an example to be open by default. This feature is useful for error explanations, where we would like for this to be the case.

### DIFF
--- a/Manual/Meta/Example.lean
+++ b/Manual/Meta/Example.lean
@@ -16,19 +16,23 @@ open Lean Elab
 
 namespace Manual
 
-def Block.example (name : Option String) : Block where
+def Block.example (name : Option String) (opened : Bool) : Block where
   name := `Manual.example
-  data := ToJson.toJson (name, (none : Option Tag))
+  data := ToJson.toJson (name, opened, (none : Option Tag))
 
 structure ExampleConfig where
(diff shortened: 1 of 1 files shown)


Manual/Meta/Example.lean
```diff
@@ -16,19 +16,23 @@ open Lean Elab
...
   keep : Bool := false
+  opened : Bool := false
 
 
 def ExampleConfig.parse [Monad m] [MonadInfoTree m] [MonadLiftT CoreM m] [MonadEnv m] [MonadError m] [MonadFileMap m] : ArgParse m ExampleConfig :=
-  ExampleConfig.mk <$> .positional `description .inlinesString <*> .named `tag .string true <*> (.named `keep .bool true <&> (·.getD false))
+  ExampleConfig.mk <$> .positional `description .inlinesString
+                   <*> .named `tag .string true
+                   <*> (.named `keep .bool true <&> (·.getD false))
+                   <*> (.named `opened .bool true <&> (·.getD false))
```
david-christiansen:
```suggestion
                   <*> (.named `open .bool true <&> (·.getD false))
```
I think `opened` implies that it has been opened, but here it's already open from the start.

PR #480, opened 2025-06-24
feat: add error explanation preprocessor

jrr6:
This PR adds a preprocessor for elaborating error explanation code blocks in batches based on their import sets.

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -27,7 +27,8 @@ lean_lib Manual where
 
 def figureDir : FilePath := "figures"
 def figureOutDir : FilePath := "static/figures"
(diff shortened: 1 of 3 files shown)


lakefile.lean
```diff
@@ -79,7 +85,142 @@ target figures : Array FilePath := do
 
     pure srcInfo
 
+section ExplanationPreprocessing
+open Lean Meta
```
david-christiansen, line 116: A comment here that contains an overall explanation of what this section does and the role played by each definition would be very valuable here.

jrr6: Added!

PR #481, opened 2025-06-24
feat: add error explanation elaboration and manual page

jrr6:
This PR adds elaboration and a manual page for error explanations.

### DIFF
--- a/.vale/scripts/rewrite_html.py
+++ b/.vale/scripts/rewrite_html.py
@@ -30,6 +30,13 @@ def process_html_file(filepath, output_filepath):
         elif code_tag.attrs and 'class' in code_tag.attrs and 'hl' in code_tag['class'] and 'lean' in code_tag['class']:
             code_tag.decompose()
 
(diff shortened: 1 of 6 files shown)


david-christiansen:
I think that we should just disable the prose linter for error explanations. The feedback loop is way too long for it to be worth it. If we want one for them, then we should configure it upstream, in the `lean4` CI.

david-christiansen:
I can just make the prose linter ignore any HTML that contains an element with class `error-example-container`

david-christiansen:
You haven't enabled the "allow maintainers to edit PR" flag, so I can't push it directly, but I do have a fix for this. Can you enable the flag?

jrr6:
> You haven't enabled the "allow maintainers to edit PR" flag, so I can't push it directly, but I do have a fix for this. Can you enable the flag?

I think you should be able to commit directly to the `jrr6/explanations-remainder` branch and have the changes show up here.

david-christiansen:
I managed to confuse myself about the role the `/` was playing here!

PR #488, opened 2025-06-26
feat: update nightly-testing automatically when possible

david-christiansen:


david-christiansen:
Looks like I can't test `workflow-dispatch` unless it's in `main`...

PR #492, opened 2025-06-27
Fix plusTwo bug in Intro

nguyenvukhang:


david-christiansen:
I just tested, and it does indeed work.

Perhaps you're in a context where automatic implicits are turned off? I believe that is the default setting in the mathlib project template in 4.21.0-rc3.

nguyenvukhang:
Indeed, it turns out that automatic implicits are off in my environment. Thanks for your patience.

PR #508, opened 2025-06-30
chore: v4.22-rc1 release

david-christiansen:
This PR squashes all the bump PRs in nightly-testing and rebases the commits with substantive content.

Next development cycle, this won't be needed. Once I've done some more work to verify that no content is missing here, I'll do a reset on `nightly-testing` and document a better workflow for the next cycle.


david-christiansen:
OK, I've manually inspected the diffs between this and `main` and `nightly-testing` and we're good.

PR #512, opened 2025-07-01
doc: swap sides of equality in grind

adomani:
The text below the `example` refers to the hypothesis `p (q x) = 7` and to `7` as the "right-hand-side".  The proposed order also satisfies the "LHS is more complicated than the RHS" heuristic.


robsimmons:
Thank you for this, and sincere apologies for letting it languish! Closing in favor of #841, which updates the fix to the current manual (and fixes a bug introduced by the ravages of time and refactoring to boot).

adomani:
Great: I'm happy to see the improvements!

PR #514, opened 2025-07-01
feat: add module system appendix

Kha:


Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+open Verso.Genre.Manual.InlineLean
+
+#doc (Manual) "The Module System" =>
+%%%
+%%%
+
+The module system is an experimental Lean 4 feature allowing for finer-grained control over what information is exported from Lean modules and imported into other modules, improving average build times, executable sizes, API design options, and more.
+
+The module system is activated by prefixing a Lean file with the `module` keyword.
+At the time of writing, it is additionally guarded by the `experimenta
```
nomeata:
```suggestion
At the time of writing, it is additionally guarded by the `experimental
```

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+```
+module
+
+def priv : Nat := 0
+
+public theorem priv_eq_priv : priv = priv := rfl
+```
+
+`public section` can be used to switch the default scope for declarations, with `private` locally negating it.
+This is mostly intended to ease porting and avoiding merge conflicts during it.
```
nomeata:
```suggestion
This is mostly intended to ease porting and avoiding merge conflicts in the process.
```

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+
+def myInternalHelper (x : Nat) : String := ...
+
+public instance : ToString Nat where
+  toString x := private myInternalHelper x
+```
+
+## Import Visibility
+
+The basic form `public import M` makes the public scope of `M` available in the public scope of the current module and discards the private scope.
```
nomeata:
```suggestion
The basic form `public import M` makes the public scope of `M` available in the public scope of the current module. The private scope of `M` is discarded.
```

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+public theorem f_eq_zero : f = 0 :=
+  -- may access body of `f` imported into the private scope
+  rfl
+```
+Note that the imported private scope includes private `import`s of `M`, including nested `import all`s that then are interpreted likewise.
+That is, the set of private scopes accessible to the current module is the transitive closure of `import all` declarations.
+Note also that `import all` is more powerful than the basic `import` before the module system as it allows for accessing imported private definitions directly by name, as if they were defined in the current module.
+Thus another use case of `import all` is to make declarations available that need to be used in multiple modules but should not leak outside the current library.
+
+`public import all M` behaves like `public import M` followed by `import all M`, i.e. the `all` modifier becomes irrelevant for downstream modules.
```
nomeata, line 122: I wonder if this is really helpful (because it looks like the `all` is made public), or if we should require writing 
```
public import M
import all M
```
instead, unless this is a common idiom.

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+import Manual.Meta
+
+open Verso.Genre Manual
+open Verso.Genre.Manual.InlineLean
+
+#doc (Manual) "The Module System" =>
+%%%
+%%%
+
+The module system is an experimental Lean 4 feature allowing for finer-grained control over what information is exported from Lean modules and imported into other modules, improving average build times, executable sizes, API design options, and more.
```
david-christiansen:
```suggestion
The module system is an experimental feature allowing for finer-grained control over what information is exported from Lean modules and imported into other modules, improving average build times, executable sizes, API design options, and more.
```
The manual refers to the system as just "Lean", because we expect that as we grow, former Lean 3 users will be a tiny minority, so the distinction isn't so important. And in this context, I think it's clear that it's a Lean feature, so the whole name can be dropped.

We also try to avoid single-sentence paragraphs, and this one is a bit long anyway. How about:
```suggestion
The module system is an experimental feature that allows for more fine-grained control over what information is exported from, and imported into, modules.
This can improve build times, reduce the sizes of compiled executables, provide more API design options, and more.
```

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+
+The module system is activated by prefixing a Lean file with the `module` keyword.
+At the time of writing, it is additionally guarded by the `experimenta
+.module` option that has to be set to `true` in the project's Lakefile.
```
david-christiansen:
```suggestion
.module` option that has to be set to `true` in the project's Lake configuration file.
```
@tydeu prefers this terminology

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+`module`s can only import other `module`s so adoption has to be done top-down.
+Non-`module`s can import `module`s and will ignore all module system-specific annotations.
```
david-christiansen:
```suggestion
Modules that opt in to the module system with the `module` keyword can only import other modules that themselves use the module system, so adoption has to be done top-down.
Modules that do not use the module system can import modules that do, and will ignore all module-system-specific annotations.
```

I think that using only `code font` to distinguish between Lean modules that do or do not use the module system risks being a bit unclear.

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+
+# Visibility
+
+The main distinction the module system introduces is between the *public scope* that contains all information visible to other modules via `import` and the *private scope* that is not imported by default.
```
david-christiansen:
```suggestion
The main distinction the module system introduces is between the {deftech}_public scope_ that contains all information visible to other modules via `import` and the {deftech}_private scope_ that is not imported by default.
```
This annotates them as technical terminology, which adds them to the quick jump box and makes them into link targets for use sites.

Manual/ModuleSystem.lean
david-christiansen: Or even:
```suggestion
This is mostly intended to ease porting while avoiding merge conflicts.
```

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+
+public theorem priv_eq_priv : priv = priv := rfl
+```
+
+`public section` can be used to switch the default scope for declarations, with `private` locally negating it.
+This is mostly intended to ease porting and avoiding merge conflicts during it.
+
+Marking a declaration as public at minimum makes its "signature", i.e. its name and type, visible.
+Some specific declarations/terms may still put other parts in the private scope as follows:
+* `by` used in the public scope on a proposition puts the resulting proof in the private scope (by wrapping it in a public helper theorem).
```
david-christiansen:
```suggestion
* `by` used in the public scope to prove a proposition puts the resulting proof in the private scope (by wrapping it in a public helper theorem).
```
The original phrasing took two passes for me to understand

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+* `def` puts its body in the private scope by default, thus it cannot be unfolded when used in the public scope.
```
david-christiansen:
```suggestion
* `def` puts its body in the private scope by default. The defined constant cannot be unfolded when used in the public scope.
```
Comma splice. Also, "it" should refer to the defined constant, but here it sounded more like the keyword/syntactic for `def` itself.

Manual/ModuleSystem.lean
david-christiansen: Also, I'd use `{tech}[public scope]` here to make it link to the definition of the term, as it's been a bit of text.

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+def myInternalHelper (x : Nat) : String := ...
+
+public instance : ToString Nat where
+  toString x := private myInternalHelper x
+```
+
+## Import Visibility
+
+The basic form `public import M` makes the public scope of `M` available in the public scope of the current module and discards the private scope.
+Without `public`, the public scope of `M` is instead imported into the private scope.
```
david-christiansen:
```suggestion
Without `public`, the public scope of `M` is instead imported into the {tech}[private scope].
```

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+The import thus is irrelevant to downstream modules and ignored by them.
+
+`import all M` behaves like `import M` but additional imports the private scope of `M` into the private scope of the current module.
+This is only allowed if `M` and the current module live in the same library as its main purpose is to allow for separating definitions and proofs into separate modules for internal organization of a library.
```
david-christiansen: Just to be sure: is this library or package, in the Lake sense?

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+
+import all M.Defs
+
+public theorem f_eq_zero : f = 0 :=
+  -- may access body of `f` imported into the private scope
+  rfl
+```
+Note that the imported private scope includes private `import`s of `M`, including nested `import all`s that then are interpreted likewise.
+That is, the set of private scopes accessible to the current module is the transitive closure of `import all` declarations.
+Note also that `import all` is more powerful than the basic `import` before the module system as it allows for accessing imported private definitions directly by name, as if they were defined in the current module.
```
david-christiansen:
```suggestion

The module system's `import all` is more powerful than `import` without the module system.
It makes imported private definitions accessible directly by name, as if they were defined in the current module.
```

I think this should be a new paragraph and two sentences.

Manual/ModuleSystem.lean
david-christiansen: Even if it's common, I have the same gut feeling when seeing presentations and reading the text. It feels like a footgun.

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+# The `meta` Phase
+
+When it comes to actual code execution, there is no point to a definition without a body.
+Thus, in order to eagerly know what definitions *might* be executed at compile time and so need to be available including their bodies (in some executable shape), any definition used as an entry point to compile-time execution has to be tagged with the new `meta` modifier[^meta3].
+This is automatically done in built-in metaprogramming syntax such as `syntax`, `macro`, and `elab` but may need to be done explicitly when manually applying metaprogramming attributes such as `@[app_delab]`.
+
+[^meta3]: Semantically unrelated to the modifier of the same name in Lean 3.
+
+A `meta` definition may access (and thus invoke) any `meta` or non-`meta` definition of the current module.
+For accessing imported definitions, the definition must either have been marked as `meta` when it was declared or the import must be marked as such (`meta import` when the accessing definition is in the private scope and `public meta import` otherwise).
```
david-christiansen, line 133: A question I have when reading this is whether `meta import` is a superset of `import` or whether it's orthogonal.

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+That is, the set of private scopes accessible to the current module is the transitive closure of `import all` declarations.
+Note also that `import all` is more powerful than the basic `import` before the module system as it allows for accessing imported private definitions directly by name, as if they were defined in the current module.
+Thus another use case of `import all` is to make declarations available that need to be used in multiple modules but should not leak outside the current library.
+
+`public import all M` behaves like `public import M` followed by `import all M`, i.e. the `all` modifier becomes irrelevant for downstream modules.
+
+# The `meta` Phase
+
+When it comes to actual code execution, there is no point to a definition without a body.
+Thus, in order to eagerly know what definitions *might* be executed at compile time and so need to be available including their bodies (in some executable shape), any definition used as an entry point to compile-time execution has to be tagged with the new `meta` modifier[^meta3].
```
david-christiansen:
```suggestion
Thus, in order to eagerly know what definitions _might_ be executed at compile time and so need to be available including their bodies (in some executable shape), any definition used as an entry point to compile-time execution has to be tagged with the new `meta` modifier[^meta3].
```
For stylistic consistency with the rest of the manual

Manual/ModuleSystem.lean
```diff
@@ -0,0 +1,132 @@
...
+The following list contains common errors one might encounter when using the module system and especially porting existing files to the module system.
+
+* Unknown constant: check whether you might be trying to access a private definition in the public scope.
+  If so, you might want to make the current declaration private as well or otherwise enter the private scope such as through `private` on a field or `by` for a proof.
+  TODO: improve error message.
+* Definitional equality errors, especially after porting: you are likely missing an `@[expose]` on a definition or alternatively, if imported, an `import all`.
+  Prefer the former if anyone outside your library might feasible require the same access.
+  `#reduce` and/or `trace.Meta.isDefEq` can help with finding the blocking definition.
+  You might also see this as a kernel error when a tactic directly emits proof terms referencing specific declarations without going through the elaborator, such as for proof by reflection.
+  In this case, there is no readily available trace for debugging; consider using `@[expose] section`s generously on the closure of relevant modules.
```
david-christiansen:
```suggestion
: Unknown constant
  Check whether you might be trying to access a private definition in the public scope.
  If so, you might want to make the current declaration private as well or otherwise enter the private scope such as through `private` on a field or `by` for a proof.
  TODO: improve error message.
: Definitional equality errors, especially after porting
  You are likely missing an {attr}`expose` attribute on a definition or alternatively, if imported, an `import all`.
  Prefer the former if anyone outside your library might feasible require the same access.
  {keywordOf Lean.reduceCmd}`#reduce` and/or {option}`trace.Meta.isDefEq` can help with finding the blocking definition.
  You might also see this as a kernel error when a tactic directly emits proof terms referencing specific declarations without going through the elaborator, such as for proof by reflection.
  In this case, there is no readily available trace for debugging; consider using `@[expose] section`s generously on the closure of relevant modules.
```

david-christiansen: There should also be hyperlinks for `import all` and `@[expose] section`, but I can add those - there should be a syntax definition for module system headers and special sections that can have the grammar that these link to.

david-christiansen:
It looks like a GH Action we depended on got deleted. Rebase on `main` to get a fixed CI.

Manual/ModuleSystem.lean
Kha: I had exactly the same thought after writing the paragraph, let's do it!

Manual/ModuleSystem.lean
Kha: I feel we need to come up with *some* succinct terminology. I'm even inclined to reserve the non-keyword term "module" to `module`s going forward but it's probably a bit too early for that.

Manual/ModuleSystem.lean
Kha: Thanks!

Manual/ModuleSystem.lean
Kha: Currently it's module name root, but that is subject to change. So "library" is close enough for now.

Manual/ModuleSystem.lean
Kha: I'm not completely sure yet to be honest; it is currently implemented as a superset. But a `public meta import` would not necessarily have imply `public import` if the imported declarations are used only in the *bodies* of public declarations.

Manual/ModuleSystem.lean
david-christiansen: I think that's a great idea, but it does require a larger change in the rest of the manual.

How about we stick to the awkward phrasing while this is an appendix, and then update our terminological conventions when it becomes something for wider use?

Manual/ModuleSystem.lean
david-christiansen: I'd rather have the text be accurate if possible:
```suggestion
This is only allowed if `M` and the current module have the same module name root, as its main purpose is to allow for separating definitions and proofs into separate modules for internal organization of a library.
```

Manual/ModuleSystem.lean
david-christiansen: I would think that my imports for purposes of metaprogramming/code generation should be reasonably distinct from my imports for purposes of generated run-time code, so that I could depend on a heavyweight library for proof automation but distribute an executable that's independent of it.

Manual/ModuleSystem.lean
david-christiansen: I'd rather have the text be accurate if possible:
```suggestion
This is only allowed if `M` and the current module have the same module name root, as its main purpose is to allow for separating definitions and proofs into separate modules for internal organization of a library.
```

Manual/ModuleSystem.lean
Kha: Ah you are right of course, we need to change that for the executable size goal

PR #522, opened 2025-07-04
chore: bump toolchain to v4.22.0-rc3

kim-em:


Manual/Releases/v4_21_0.lean
```diff
@@ -12,13 +12,14 @@ open Manual
 open Verso.Genre
 
 
-#doc (Manual) "Lean 4.21.0 (2025-06-30)" =>
+#doc (Manual) "Lean 4.21.0-rc3" =>
```
david-christiansen: This seems to contain commits that it shouldn't...

Manual/Releases.lean
```diff
@@ -54,8 +53,6 @@ the differences with the previous version and upgrade your projects.
 
 {include 0 Manual.Releases.«v4_21_0»}
 
-{include 0 Manual.Releases.«v4_20_1»}
```
david-christiansen: I don't think we want to delete this

Manual/ErrorExplanations.lean
```diff
@@ -14,7 +14,7 @@ namespace Manual
 set_option pp.rawOnError true
 set_option guard_msgs.diff true
 
-set_option manual.requireErrorExplanations true
+set_option manual.requireErrorExplanations false
```
david-christiansen: This should always be true for releases

Manual.lean
```diff
@@ -330,21 +330,6 @@ Std.DTreeMap
 Std.TreeSet
 ```
 
-
```
david-christiansen: These should be here! Otherwise the progress bar is inaccurate.

Issue #434, opened 2025-05-06
Is there a downloadable PDF version of the reference manual?


euprunin:
Is there a downloadable PDF version of the reference manual?

If not, what’s the recommended way to generate one?

david-christiansen:
`lake exe generate-manual --with-html-single` should do the trick. Be warned: it's big, and the navigation features are really designed for the multi-page mode. I can't make any browser on my system reliably produce a PDF from it (but I'd merge the PR that did, all else equal)

I-Al-Istannen:
The Firefox-generated PDF can be repaired with `mutool clean "The Lean Language Reference.pdf" out.pdf`, but that sounds like a Firefox bug.

AaronNGray:
@david-christiansen Is there anyway chance this and making Verso output all the main four LEAN PDF manuals properly and  and maybe one or two interesting projects could be given as paid Summer of Code project for a student. LEAN could really do with proper manuals, and maybe even get them into hard copy printable form for sale ?

Oh we missed the GSoC 2026 deadline by a long way :(

PR #527, opened 2025-07-04
fix: make printed page look somewhat decent

Garmelon:
The nav bar took up a lot of space even though it wasn't displayed. I also got rid of the header while I was there.

Some code snippets were printed with a scroll bar instead of wrapping them. Also, footnotes can't float in print note and just looked broken, so I redesigned them as in-line info boxes. They still look a little wonky, but they're a lot better than overlapping digits onto text.

These changes are the result of my experiments on zulip: [#new members > Printable form of language reference? @ 💬](https://leanprover.zulipchat.com/#narrow/channel/113489-new-members/topic/Printable.20form.20of.20language.20reference.3F/near/527227988)

They should help with #434 if one uses a browser to convert the single-page manual to a PDF.

### DIFF
--- a/static/print.css
+++ b/static/print.css
@@ -19,6 +19,9 @@
     pre {
         white-space: pre-wrap;
     }
+    pre.hl.lean {
+        white-space: pre-wrap;
+    }
     .hl.lean.block {
         white-space: pre-wrap;
     }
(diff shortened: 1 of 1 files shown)


david-christiansen:
Let me know when you're ready for a merge - looks like you're still working on it, so I'll hold off for now.

david-christiansen:
And thanks, BTW!

PR #533, opened 2025-07-09
Corrected link for custom operator priorities.

RichardKelley:
In the section [Custom Operators](https://lean-lang.org/doc/reference/latest/Notations-and-Macros/Custom-Operators/), the discussion of operator priority currently has a `{tech}[priority]` link to the definition of Task priority in the section [Tasks and Threads](https://lean-lang.org/doc/reference/latest/IO/Tasks-and-Threads/). But the productions on the Custom Operators page refer to the nonterminal `prio`, which is defined in the section [Instance Declarations](https://lean-lang.org/doc/reference/latest//Type-Classes/Instance-Declarations/#instance-priorities). The `{tech}` link on the Custom Operator page should point to the location where `prio` is defined. 

This PR changes the `deftech` for both instance priorities and thread priorities so that the link on the Custom Operators page points to the correct section of the manual.

### DIFF
--- a/Manual/Classes/InstanceDecls.lean
+++ b/Manual/Classes/InstanceDecls.lean
@@ -292,7 +292,7 @@ instance instDecidableEqStringList : DecidableEq StringList
 tag := "instance-priorities"
 %%%
 
-Instances may be assigned {deftech}_priorities_.
+Instances may be assigned a {deftech}_priority_.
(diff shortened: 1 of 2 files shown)


jcreedcmu:
This has been inactive in a draft state for a while so we'll close now. Feel free to reopen if you still want to work on this!

PR #537, opened 2025-07-17
doc: propagate comments in the example grind propagator

YaelDillies:
There was one useful comment in one of the four branches, but none in the other three. This PR adds them.


YaelDillies:
The missing comments were added in #554.

PR #559, opened 2025-08-07
chore: show build job timing in CI summary

david-christiansen:


david-christiansen:
Looks like this needs the next release :)

PR #566, opened 2025-08-14
feat: list `mvcgen` and stateful proof mode tactics in tactic reference

sgraf812:
This PR lists `mvcgen` and the stateful proof mode tactics in tactic reference.

### DIFF
--- a/Manual/Meta/ErrorExplanation.lean
+++ b/Manual/Meta/ErrorExplanation.lean
@@ -13,7 +13,8 @@ import Lean.ErrorExplanations
 import PreprocessedExplanations
 
 open Lean Elab
-open Verso.ArgParse
+open Verso (ArgParse)
+open Verso.ArgParse (parseThe)
 open Verso.Doc Elab
 open Verso.Genre.Manual Markdown InlineLean
 open SubVerso.Highlighting
--- a/Manual/Meta/LexedText.lean
+++ b/Manual/Meta/LexedText.lean
@@ -134,13 +134,12 @@ def Inline.c (value : LexedText) : Inline where
 
 def lexedText := ()
 
-@[code_block_expander c]
-def c : CodeBlockExpander
-  | args, str => do
-    ArgParse.done.run args
+@[code_block]
+def c : CodeBlockExpanderOf Unit
+  | (), str => do
     let codeStr := str.getString
     let toks ← LexedText.highlight hlC codeStr
-    pure #[← ``(Block.other (Block.c $(quote toks)) #[Block.code $(quote codeStr)])]
+    ``(Block.other (Block.c $(quote toks)) #[Block.code $(quote codeStr)])
 
 open Verso.Output Html in
 open Verso.Doc.Html in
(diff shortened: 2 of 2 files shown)


Manual/Tactics/Reference.lean
```diff
@@ -660,3 +660,99 @@ tag := "tactic-ref-other"
...
+:::
+
+:::tactic "mexists"
+:::
+
+{docstring Lean.Parser.Tactic.mpureIntroMacro}
+
+-- The followign produces an error: "Failed to parse docstring as Markdown"
+-- :::tactic "mpure_intro"
+-- :::
```
sgraf812: Not sure what is causing the error here; referencing the relevant docstring seems to work?

david-christiansen: It's because there's multiple tactics defined with this name, not registered as aliases:
```
#[{ internalName := `Lean.Elab.Tactic.Do.ProofMode.tacticMpure_intro,
    userName := "mpure_intro",
    tags := Std.TreeSet.ofList [],
    docString := none,
    extensionDocs := #[] },
  { internalName := `Lean.Parser.Tactic.mpureIntro,
    userName := "mpure_intro",
    tags := Std.TreeSet.ofList [],
    docString := some "`mpure_intro` operates on a stateful `Std.Do.SPred` goal of the form `P ⊢ₛ ⌜φ⌝`.\nIt leaves the stateful proof mode (thereby discarding `P`), leaving the regular goal `φ`.\n```lean\ntheorem simple : ⊢ₛ (⌜True⌝ : SPred σs) := by\n  mpure_intro\n  exact True.intro\n```\n",
    extensionDocs := #[] },
  { internalName := `Lean.Parser.Tactic.mpureIntroMacro,
    userName := "mpure_intro",
    tags := Std.TreeSet.ofList [],
    docString := some "`mpure_intro` operates on a stateful `Std.Do.SPred` goal of the form `P ⊢ₛ ⌜φ⌝`.\nIt leaves the stateful proof mode (thereby discarding `P`), leaving the regular goal `φ`.\n```lean\ntheorem simple : ⊢ₛ (⌜True⌝ : SPred σs) := by\n  mpure_intro\n  exact True.intro\n```\n",
    extensionDocs := #[] }]
```
I'll make the error message better and the whole thing more robust, but we should fix this upstream too.

david-christiansen (approved):

Manual/Tactics/Reference.lean
sgraf812: Thanks for debugging this! The first entry in your list should vanish after https://github.com/leanprover/lean4/pull/9938. After that `mpure_intro` should be no different to the other tactics, in that they follow the general `bv_decide` pattern of "(1) define an early available `m*Macro` with `priority := low` in `Init` that fails saying you should import `Std.Tactic.Do`, (2) define the actual `m*` macro in `Std.Tactic.Do` that overrides `m*Macro` and has a builtin elaborator referring to stuff from `Std`".

PR #572, opened 2025-08-17
feat: add `toTex` support for `Manual.keywordOf`

kpadmasola:
This PR adds TeX rendering support for `Manual.keywordOf`.

Relates to https://github.com/leanprover/reference-manual/issues/434

### DIFF
--- a/Manual/Meta/Syntax.lean
+++ b/Manual/Meta/Syntax.lean
@@ -125,10 +125,7 @@ def keywordOf : RoleExpander
 def keywordOf.descr : InlineDescr where
   traverse _ _ _ := do
     pure none
-  toTeX :=
-    some <| fun goI _id _ content => do
-      pure <| .seq <| ← content.mapM fun b => do
-        pure <| .seq #[← goI b]
+  toTeX := some fun goI _ _ content => content.mapM goI
   toHtml :=
     open Verso.Output Html in
     some <| fun goI _ info content => do
--- a/Manual/Meta/Syntax.lean
+++ b/Manual/Meta/Syntax.lean
@@ -125,7 +125,10 @@ def keywordOf : RoleExpander
 def keywordOf.descr : InlineDescr where
   traverse _ _ _ := do
     pure none
-  toTeX := none
+  toTeX :=
+    some <| fun go _id _ content => do
+      pure <| .seq <| ← content.mapM fun b => do
+        pure <| .seq #[← go b, .raw "\n"]
   toHtml :=
     open Verso.Output Html in
     some <| fun goI _ info content => do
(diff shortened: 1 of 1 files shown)


Manual/Meta/Syntax.lean
```diff
@@ -125,7 +125,10 @@ def keywordOf : RoleExpander
 def keywordOf.descr : InlineDescr where
   traverse _ _ _ := do
     pure none
-  toTeX := none
+  toTeX :=
+    some <| fun go _id _ content => do
+      pure <| .seq <| ← content.mapM fun b => do
+        pure <| .seq #[← go b, .raw "\n"]
```
david-christiansen: Why does this need a newline after it?

kpadmasola: Since this is supposed to be rendered inline, it should not have a newline. Removed.

Manual/Meta/Syntax.lean
```diff
@@ -125,7 +125,10 @@ def keywordOf : RoleExpander
 def keywordOf.descr : InlineDescr where
   traverse _ _ _ := do
     pure none
-  toTeX := none
+  toTeX :=
+    some <| fun goI _id _ content => do
+      pure <| .seq <| ← content.mapM fun b => do
+        pure <| .seq #[← goI b]
```
david-christiansen: This should be equivalent, but simpler:
```suggestion
  toTeX := some fun goI _ _ content => content.mapM goI
```

kpadmasola: thanks, incorporated the suggestion.

david-christiansen:
Hmm, it certainly builds with the results of #576 locally, and it would really surprise me if CI let us commit something to `main` that didn't build. Try `lake clean` before building, perhaps?

kpadmasola:
@david-christiansen I figured out why I was getting the error. I had done a `lake update` to get the new verso version, but forgot to pull the latest changes from `main` branch in this repo. I remember checking that my fork was up to date, but I must have forgotten to pull the changes locally. My mistake :-)

david-christiansen:
Good news!

The thing to do is not to put placeholder text in the `toTeX` implementations one-by-one, but rather get Verso to insert it when there is no such implementation. That way, it becomes obvious what needs to be done in order to build a given document.

kpadmasola:
@david-christiansen After manually fixing the errors in the generated `main.tex` file, I was able to generate `main.pdf` by running `lualatex main.tex`. I have collected the issues that we need to fix. The errors fall into about 7 or 8 categories. The fixes will be on the Verso side I suppose, although I haven't identified the code level changes yet.

I'll raise the verso PR now. Next I'll try to identify the code level changes needed to fix TeX compilation errors. If you have any suggestions about where I should look, please let me know.

Here's the generated [main.pdf](https://github.com/user-attachments/files/21992325/main.pdf). I haven't gone through it yet. Initially we should make it work, i.e., no TeX errors, and then make it make it look good.

jcreedcmu:
This ended up getting implemented in a different way, but thanks for the contribution!

PR #573, opened 2025-08-18
feat: module system porting guide

Kha:


Manual/ModuleSystem.lean
```diff
@@ -163,3 +163,14 @@ The following list contains common errors one might encounter when using the mod
   {keywordOf Lean.reduceCmd}`#reduce` and/or {option}`trace.Meta.isDefEq` can help with finding the blocking definition.
   You might also see this as a kernel error when a tactic directly emits proof terms referencing specific declarations without going through the elaborator, such as for proof by reflection.
   In this case, there is no readily available trace for debugging; consider using `@[expose] section`s generously on the closure of relevant modules.
+
+## Recipe for Porting Existing Files
+
+Start by enabling the module system throughout all files with minimal breaking changes:
```
david-christiansen, line 169: I think a `lakefile.toml` and perhaps `lakefile.lean` snippet that shows how to enable the option on the command line would be good to have here

Kha: I think the small technical hurdle while the experimental option even still exists is not completely without merit

PR #582, opened 2025-09-08
feat: clarify cyclic data detection

nomeata:
The previous wording was imprecise if one considers `IO.Ref` and `Dynamic` etc, so refine.


Manual/Runtime.lean
```diff
@@ -110,7 +110,7 @@ Nevertheless, multi-threaded code requires that reference count updates are sync
 To reduce this overhead, Lean values are partitioned into those which are reachable from multiple threads and those which are not.
 Single-threaded reference counts can be updated much faster than multi-threaded reference counts, and many values are accessed only on a single thread.
 Together, these techniques greatly reduce the performance overhead of reference counting.
-Because Lean cannot create cyclic data, no technique is needed to detect it.
+The verifiable fragment of Lean cannot create cyclic data anyways, so the Lean runtime does not have a technique to detect it.
```
david-christiansen: I think this fits the overall style and register of the paragraph a bit better:
```suggestion
Because the verifiable fragment of Lean cannot create cyclic data, the Lean runtime does not have a technique to detect it.
```

PR #585, opened 2025-09-11
chore: add paragraph about `grind?` and `grind only`

kim-em:


Manual/Grind.lean
```diff
@@ -151,6 +151,15 @@ In particular, it presents equivalence classes of terms that it has determined t
 The two largest classes are shown as `True propositions` and `False propositions`, listing every literal currently known to be provable or refutable.
 Inspect these lists to spot missing facts or contradictory assumptions.
 
+# Minimizing `grind` calls
+
+By replacing a successful {tactic}`grind` tactic invocation with {tactic}`grind?`,
+you will see a message with associated code action
+that will replace the {tactic}`grind?` with a `grind only [...]` tactic, showing exactly the theorems {tactic}`grind` used.
```
david-christiansen: Small details:
```suggestion
The `grind only [...]` tactic invokes {tactic}`grind` with a limited set of theorems, which can improve performance.
Calls to `grind only` can be conveniently constructed using {tactic}`grind?`, which automatically records the theorems used by {tactic}`grind` and suggests a suitable `grind only`.
```

Here's the nitpicky reason: `grind?` works fine without having written `grind` first, so users don't  need to replace a successful `grind` with `grind?`. And I think it's good to explain `grind only` before using it here (and I need to get the right markup working to refer to that, it seems).

Manual/Grind.lean
```diff
@@ -151,6 +151,15 @@ In particular, it presents equivalence classes of terms that it has determined t
...
+
+This theorems will typically include a symbol prefix such as `=`, `←`, or `→`, indicating the
```
david-christiansen:
```suggestion
These theorems will typically include a symbol prefix such as `=`, `←`, or `→`, indicating the
```

Manual/Grind.lean
```diff
@@ -151,6 +151,15 @@ In particular, it presents equivalence classes of terms that it has determined t
...
+pattern that triggered the instantiation. See the section on E-matching for details.
```
david-christiansen: This will add a link:
```suggestion
pattern that triggered the instantiation. See the {ref "e-matching"}[section on E-matching] for details.
```

PR #586, opened 2025-09-11
feat: section on annotating libraries for grind

kim-em:


david-christiansen:
Hmm, looking at the text, it isn't quite ready. I'll give it a pass and see if I can get it finished up (it includes things like reader-visible "this is unfinished" notes).

PR #595, opened 2025-09-19
chore: use extraHead to include static CSS and JS in the manual

robsimmons:
This is a no-op change for the current verso, but its necessary for the reference manual to continue working when leanprover/verso#530 gets merged, because that PR moves `extraJs` and `extraCss` to including inline js/css (making it parallel to other configuration objects in verso) and requires static js and css assets to be included, for now at least, via extraHead configuration.

### DIFF
--- a/Main.lean
+++ b/Main.lean
@@ -15,25 +15,28 @@ def plausible := {{
     <script defer="defer" data-domain="lean-lang.org" src="https://plausible.io/js/script.outbound-links.js"></script>
   }}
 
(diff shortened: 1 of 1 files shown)


robsimmons:
@david-christiansen I'm going to preemptively merge this, but let me know if you want to have a conversation about whether we want to ultimately do this via extraHead or via something a little more like what we had before, since obviously this other way of doing things *was* being used somewhere

PR #596, opened 2025-09-19
feat: precise spec for natural numbers and scientific notation literals

robsimmons:
This PR makes the lexical specification of natural numbers and scientific notation literals more precise.

### DIFF
--- a/Manual/Terms.lean
+++ b/Manual/Terms.lean
@@ -985,7 +985,7 @@ Terms such as {lean}`-5` consist of a prefix negation (which can be overloaded v
 
 ## Scientific Numbers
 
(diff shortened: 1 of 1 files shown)


Manual/Terms.lean
```diff
@@ -913,7 +913,7 @@ end
 ```
 ::::
 
-# Literals
+# Numeric Literals
```
robsimmons, line 916: Am I going to screw around with anchor refs or worse by changing this? The current title is misleading, as string literals are covered in a totally different section.

robsimmons:
Note: I ended up not going with a BNF-style notation here because I couldn't find any other BNF representations of *tokens* (as opposed to terms) in the language reference.

PR #598, opened 2025-09-23
feat: add a tutorial chapter for `mvcgen`

sgraf812:


david-christiansen:
@sgraf812 I made a number of updates, and it's ready for you to check them. Thanks for your patience here!

sgraf812:
Thanks, David! I gave it a full read and made some mostly minor edits. One edit wasn't minor; I un-TODO'd the `mvcgen invariants?` paragraph because it is fully supported since 4.24. In the HTML view, the suggestion it reports was hidden under the tooltip for the proof state, so I decided it's simpler to use `#guard_msgs (info)`. Wouldn't mind if you double-checked that for a better solution. 

Otherwise LGTM :) Thank you for contributing the nice technical introduction sections!

Manual/VCGen.lean
```diff
@@ -257,13 +257,21 @@ end
...
+Note that the form `mvcgen invariants?` will suggest an initial invariant using {name}`Invariant.withEarlyReturn`, so there is no need to memorize the exact syntax for specifying invariants:
+```lean
+/--
+info: Try this:
+  [apply] invariants
+  ·
+    Invariant.withEarlyReturn (onReturn := fun r letMuts => ⌜l.Nodup ∧ (r = true ↔ l.Nodup)⌝) (onContinue :=
+      fun xs letMuts => ⌜xs.prefix = [] ∧ letMuts = ∅ ∨ xs.suffix = [] ∧ l.Nodup⌝)
+-/
+#guard_msgs (info) in
```
david-christiansen: We usually show output in its own Verso block when we're not explicitly demonstrating techniques with `#guard_msgs`. Is it OK if I make that change?

sgraf812: Yes! I just didn't quite know how to do that. I found `+error`, but that wasn't quite right

Manual/VCGen.lean
```diff
@@ -55,12 +55,12 @@ open Std.Do
 
 ## Preconditions and Postconditions
 
-One style in which program specifications can be written is to provide a {deftech}_precondition_, which the program's caller is expected to ensure, and a {deftech}_postcondition_, which the program itself is expected to ensure.
-The program is correct if running it when the precondition holds always results in the postcondition holding.
+One style in which program specifications can be written is to provide a {deftech}_precondition_ $`P`, which the caller of a program `prog` is expected to ensure, and a {deftech}_postcondition_ $`Q`, which the `prog` is expected to ensure.
+The program `prog` satisfies the specification if running it when the precondition $`P` holds always results in the postcondition $`Q` holding.
```
david-christiansen: I think that if we're using `prog` as a mathematical metavariable, then it should be written in math font rather than in code font, because mono indicates to me that it's the name of a concrete, defined program.  Maybe `\mathit{prog}`?

Manual/VCGen.lean
david-christiansen: I can just do it, no prob. OK if I push some here or do you have more commits to push?

Manual/VCGen.lean
sgraf812: Great!

david-christiansen: (I'm happy to make this change if it's OK with you)

Manual/VCGen.lean
sgraf812: Yes, the lock is all yours.

PR #599, opened 2025-09-25
feat: make release notes insert parts from markdown

jcreedcmu:
Previously, we turned all headers into emphasized text regardless of level. With this change, we insert nested parts into the current verso document corresponding to the markdown structure.

### DIFF
--- a/Manual/Meta/Markdown.lean
+++ b/Manual/Meta/Markdown.lean
@@ -28,12 +28,11 @@ def Block.noVale.descr : BlockDescr where
     some <| fun _ goB _ _ content => do
       pure {{<div class="no-vale">{{← content.mapM goB}}</div>}}
 
-@[code_block_expander markdown]
-def markdown : CodeBlockExpander
-  | _args, str => do
-    let str ← parserInputString str
-    let some ast := MD4Lean.parse str
-      | throwError "Failed to parse docstring as Markdown"
-    let content ← ast.blocks.mapM <|
-      Markdown.blockFromMarkdown (handleHeaders := Markdown.strongEmphHeaders)
-    pure #[← ``(Block.other Block.noVale #[$content,*])]
+open Lean.Doc.Syntax in
+@[part_command Lean.Doc.Syntax.codeblock]
+def markdown : PartCommand
+  | `(block| ``` markdown | $txt ``` ) => do
+     let some ast := MD4Lean.parse txt.getString
+       | throwError "Failed to parse body of markdown code block"
+     _ ← ast.blocks.mapM Markdown.addPartFromMarkdown
+  | _ => Elab.throwUnsupportedSyntax


jcreedcmu:
This isn't right yet; it inserts consecutive parts more deeply nested than they ought to be, e.g. release note markdown
```
## foo
## bar
## baz
```
becomes rendered as if it was
```
## foo
### bar
#### baz
```

jcreedcmu:
I think I see what I'm supposed to do instead from `addPartsFromMarkdown`'s docstring...

jcreedcmu:
This is now, I think, correct, unless we expect permalink icons to appear while hovering over section headers in the release notes. I tentatively expected that, and I observe that they do not appear.

I think traversal is still correctly assigning fresh ids to sections arising from markdown in the release notes, because the sections appear in the TOC sidebar, and clicking on them still correctly goes to the right url-with-fragment. [I think this is where](https://github.com/leanprover/verso/blob/main/src/verso-manual/VersoManual.lean#L685-L686) permalinks get created.

Manual/Releases/v4_19_0.lean
```diff
@@ -47,11 +47,6 @@ which is added in this version via [#7366](https://github.com/leanprover/lean4/p
 All of these features can be disabled, and 'Goals accomplished!' icon can be configured in VS Code extension settings.
 See [leanprover/vscode-lean4#585](https://github.com/leanprover/vscode-lean4/pull/585) for the details.
 
-### Parallel Elaboration
```
jcreedcmu, line 50: There was another `Parallel Elaboration` section in the release notes with which this one was redundant. The other one contained #7084 with identical description, and many more items.

Manual/Releases/v4_20_0.lean
```diff
@@ -166,11 +166,6 @@ Other notable library developments in this release include:
 - Updates to `Perm` API in `Array`, `List`, and added support for `Vector`,
 - Additional lemmas for `Array`/`List`/`Vector`.
 
-### Lake
```
jcreedcmu, line 169: There was another `Lake` section in the release notes with which this one was redundant. The other one contained #7909 with almost identical description, (except it misspelled `server` as `sever`) and many more items.

nomeata: One subsection is a subsection of `Highlights` and the other one in the full list. Don't delete @eyihluyc's work 😅

Manual/Releases/v4_19_0.lean
nomeata: One subsection is a subsection of `Highlights` and the other one in the full list. Don't delete @eyihluyc's work 😅

Manual/Meta/Markdown.lean
```diff
@@ -28,12 +28,12 @@ def Block.noVale.descr : BlockDescr where
...
-  | _args, str => do
-    let str ← parserInputString str
-    let some ast := MD4Lean.parse str
-      | throwError "Failed to parse docstring as Markdown"
-    let content ← ast.blocks.mapM <|
-      Markdown.blockFromMarkdown (handleHeaders := Markdown.strongEmphHeaders)
-    pure #[← ``(Block.other Block.noVale #[$content,*])]
+@[part_command Lean.Doc.Syntax.codeblock]
+def markdown : PartCommand
+  | `(Lean.Doc.Syntax.codeblock| ``` markdown | $txt ``` ) => do
```
david-christiansen: This is matching the literal identifier `markdown`, which means that it won't  react correctly to the user's `open` statements or their own shadowing definitions. Similarly, it's better to accept arguments here, then either ignore or reject them, so that the pattern doesn't stop matching as soon as someone provides an argument.

You should instead do something like:
```suggestion
  | `(Lean.Doc.Syntax.codeblock| ``` $markdown:ident $args*| $txt ``` ) => do
    let x <- realizeGlobalConstNoOverloadsWithInfo markdown -- This actually resolves the name
    if x != ``markdown then Elab.throwUnsupportedSyntax -- If it's some other `markdown`, defer to it
    -- Now we commit to our function and can reject arguments
    for arg in args do
      let h <- m!"Remove it:" #[""] (ref? := arg)
      logErrorAt arg m!"No arguments expected{h}"
```

david-christiansen:
> This is now, I think, correct, unless we expect permalink icons to appear while hovering over section headers in the release notes. I tentatively expected that, and I observe that they do not appear.
> 
> I think traversal is still correctly assigning fresh ids to sections arising from markdown in the release notes, because the sections appear in the TOC sidebar, and clicking on them still correctly goes to the right url-with-fragment. [I think this is where](https://github.com/leanprover/verso/blob/main/src/verso-manual/VersoManual.lean#L685-L686) permalinks get created.

To get a permalink, the section has to have a stable identifier. That represents our commitment to keeping it around and keeping that name. Because release notes are basically immutable, this is easy - some mangling of the title and version will suffice. You do this by assigning the `tag` field of each section's `PartMetadata`, which will also result in them appearing in the quickjump.

Manual/Releases/v4_19_0.lean
jcreedcmu: Perish the thought! Thanks for catching my mistake. Trying to determine now why the two sections were rendered in my testing as siblings, which I now see is clearly wrong.

Manual/Meta/Markdown.lean
jcreedcmu: Thanks, makes sense, will do!

jcreedcmu:
This PR now depends on https://github.com/leanprover/verso/pull/539 to fix markdown header processing to be actually correct.

Manual/Meta/Markdown.lean
```diff
@@ -28,12 +28,17 @@ def Block.noVale.descr : BlockDescr where
...
-    let some ast := MD4Lean.parse str
-      | throwError "Failed to parse docstring as Markdown"
-    let content ← ast.blocks.mapM <|
-      Markdown.blockFromMarkdown (handleHeaders := Markdown.strongEmphHeaders)
-    pure #[← ``(Block.other Block.noVale #[$content,*])]
+@[part_command Lean.Doc.Syntax.codeblock]
+def markdown : PartCommand
+  | `(Lean.Doc.Syntax.codeblock| ``` $markdown:ident $args*| $txt ``` ) => do
+     let x <- Lean.Elab.realizeGlobalConstNoOverloadWithInfo markdown
+      if x != (by exact decl_name%) then Elab.throwUnsupportedSyntax
```
jcreedcmu: Is this `(by exact decl_name%)` tactic invocation appropriate here? As I understand it, I'm not able to literally write ``` ``markdown ``` because I want to refer to the very `def markdown` defined on line 32, right? So this is what I guessed would work instead, having seem `decl_name%` many places elsewhere. It does in fact seem to work, and using `MessageData.hint` below works to get argument deletion hints as well.

PR #601, opened 2025-09-26
feat: examples of printing axioms

robsimmons:
This PR rearranges the discussion of axioms, primarily in order to add several more examples of the use of `#print axioms <ident>`. Also links the #print axioms spec back to the docs on axioms.

Follow-up from discussion with @jcreedcmu.


Manual/Axioms.lean
```diff
@@ -176,41 +176,33 @@ Because they occur only in a proof, the compiler has no problem generating code:
...
    Quot.sound.{u} {α : Sort u}
      {r : α → α → Prop} {a b : α} :
      r a b → Quot.mk r a = Quot.mk r b
    ```
 
-Three additional axioms allow the Lean kernel to invoke code generated by the compiler, rather than using its internal reduction engine.
-This can greatly improve performance of implementations of proof by reflection.
-Rather than using these axioms directly, they are usually invoked via the {tactic}`native_decide` tactic.
-Both {name}`Lean.reduceBool` and {name}`Lean.reduceNat` contain references to {name}`Lean.trustCompiler`, which ensures that the fact that a proof trusts the correctness of the compiler is tracked.
+The first of these axiom is the Axiom of Choice; all three are discussed in the book [Theorem Proving in Lean](https://lean-lang.org/theorem_proving_in_lean4/find/?domain=Verso.Genre.Manual.section&name=axioms-and-computation).
```
jcreedcmu: as discussed elsewhere, I think I'd avoid using the exact phrase 'Axiom of Choice' here to not confuse mathematicians who think it should mean specifically the weaker (albeit related) `Classical.axiomOfChoice`

PR #616, opened 2025-10-16
fix: make the lake cache more suspicious

robsimmons:
Currently, if we change something in Verso that does not change an actual .lean file (for instance, changing a string included via `include_str`, we'll restore a lake cache that will fail to trigger recompilation, fail to reevaluate the `include_str`, and therefore will not include the updated string).

This actually happened with [this commit which was an attempted bump of v4.24.0](https://github.com/leanprover/reference-manual/commit/bdb73c25bcadace4e5f279c818ab23d76e6a5ce7).

By fully recreating `.lake` if either the lake manifest or lakefile changes, we're still subject to this problem if we have an `include_str` in the reference-manual repository, but we won't fall to it in the `verso` repository. Possibly we should go further and insist that it be entirely unchanged?


jcreedcmu:
@Kha suggested that lean4 "[uses] a simple counter in the cache name that we increment if something outside the build system's control changes", I wonder if we want to imitate that instead?

robsimmons:
Closing given that I think appropriate use of `import_dir` in Lakefiles can be an imperfect but workable solution to this problem.

PR #621, opened 2025-10-21
chore: bump to v4.25.0-rc1

david-christiansen:


lakefile.lean
```diff
@@ -8,7 +8,7 @@ import Lake
 open Lake DSL
 open System (FilePath)
 
-require verso from git "https://github.com/leanprover/verso.git"@"v4.24.0"
+require verso from git "https://github.com/leanprover/verso.git"@"main"
```
david-christiansen, line 11: `main` is intentional here to get a bugfix that isn't in the release-tagged version

PR #626, opened 2025-10-23
feat: slightly extend `meta` phase section

Kha:


Manual/ModuleSystem.lean
```diff
@@ -128,9 +128,11 @@ When it comes to actual code execution, there is no point to a definition withou
 Thus, in order to eagerly know what definitions _might_ be executed at compile time and so need to be available including their bodies (in some executable shape), any definition used as an entry point to compile-time execution has to be tagged with the new `meta` modifier.
 This is automatically done in built-in metaprogramming syntax such as `syntax`, `macro`, and `elab` but may need to be done explicitly when manually applying metaprogramming attributes such as `@[app_delab]`.
 
-A `meta` definition may access (and thus invoke) any `meta` or non-`meta` definition of the current module.
-For accessing imported definitions, the definition must either have been marked as `meta` when it was declared or the import must be marked as such (`meta import` when the accessing definition is in the private scope and `public meta import` otherwise).
-
+A `meta` definition may access (and thus invoke) other `meta` definitions only; a non-`meta` definition likewise may only access other non-`meta` definitions.
```
david-christiansen:
```suggestion
A `meta` definition may only access (and thus invoke) other `meta` definitions; a non-`meta` definition likewise may only access other non-`meta` definitions.
```
On a first read-through, the adverb came as a surprise

Manual/ModuleSystem.lean
```diff
@@ -128,9 +128,11 @@ When it comes to actual code execution, there is no point to a definition withou
...
+For imported definitions, the `meta` marker can be added after the fact using `meta import`.
+`meta import`ing a definition already in the meta phase leaves it in that phase.
+In addition, the import must be public if the imported definition may be compile-time executed outside the current module, i.e. if it is reachable from some public `meta` definition in the current module: use `public meta import` or, if already `meta`, `public import`.
```
david-christiansen, line 134:
```suggestion
In addition, the import must be public if the imported constant may be executed at compile time outside the current module.
In other words, if the imported `meta` constant is reachable from some public `meta` definition in the current module, then use `public meta import`.
If the constant is already `meta`, then `public import` is sufficient.
```

Manual/ModuleSystem.lean
```diff
@@ -128,9 +128,11 @@ When it comes to actual code execution, there is no point to a definition withou
...
+This is usually the case, unless a definition was imported solely for use in `local` metaprograms.
```
david-christiansen: Why is `local` annotated as code here?

Manual/ModuleSystem.lean
```diff
@@ -141,6 +143,13 @@ local elab "my_elab" : command => do
   ...
 ```
 
+For convenience, `meta` also implies `partial`.
+This can be overriden by giving an explicit `termination_by` metric (such as one suggested by `termination_by?`), which may be necessary when the type of the definition is not known to be `Nonempty`.
+
+As a guideline, it is usually preferable to apply `meta` as late as possible.
```
david-christiansen: I think this text is a bit ambiguous: should it be applied late in the development process when Lean complains, or should it be done via import instead of an explicit modifier? It's the latter, but I think readers might be confused.

Manual/ModuleSystem.lean
```diff
@@ -141,6 +143,13 @@ local elab "my_elab" : command => do
...
+Thus helper definitions not directly registered as metaprograms do not need to be marked if they are located in a separate module.
```
david-christiansen:
```suggestion
In particular, helper definitions that are not directly registered as metaprograms do not need to be marked if they are located in a separate module.
```
I think this is more of a "here's the important detail" sentence than an expression of causality/implication, right?

Manual/ModuleSystem.lean
```diff
@@ -141,6 +143,13 @@ local elab "my_elab" : command => do
...
+Only the final module(s) actually registering metaprograms then should use `public meta import` to import those helpers and define its metaprograms using built-in syntax like `elab`, using `meta def`, or using `meta section`.
```
david-christiansen:
```suggestion
Only the final module that actually registers a metaprogram needs the helpers as `meta` definitions.
It should use `public meta import` to import those helpers and then define its metaprograms using built-in syntax like `elab`, using `meta def`, or using `meta section`.
```
I think two sentences makes this easier to parse.

Can we motivate this? I see two reasons: that it enables re-use and independent non-meta testing of these helpers, and that things are generally just easier when not in `meta`-mode. Is there a third reason?

Manual/ModuleSystem.lean
Kha: I usually prefer "declaration" over "constant" in user-facing text, is there some canonized stance on that?

Manual/ModuleSystem.lean
Kha:
```suggestion
This is usually the case, unless a definition was imported solely for use in local metaprograms such as `local syntax/macro/elab/...`.
```

Kha: @david-christiansen Is this clearer?

Manual/ModuleSystem.lean
Kha:
```suggestion
As a guideline, it is usually preferable to keep the amount of `meta` annotations as small as possible in order to not lock reusable declarations into the meta phase unless necessary as well as to help the build system avoid more rebuilds.
Thus non-trivial amounts of code that does not contain registered metaprograms per se but that a registered metaprogram depends on should be located in a separate module and not marked `meta`.
```

Kha: @david-christiansen better?

david-christiansen: Yes!

Can we shorten the sentences a bit? E.g.
```suggestion
As a guideline, it is usually preferable to keep the amount of `meta` annotations as small as possible.
This avoids locking otherwise-reusable declarations into the meta phase and it helps the build system avoid more rebuilds.
Thus, when a metaprogram depends on other code that does not itself need to be marked `meta`, this other code should be located in a separate module and not marked `meta`.
```

Manual/ModuleSystem.lean
david-christiansen: Indeed it is.

Manual/ModuleSystem.lean
david-christiansen: No, but there should be. We typically use "declaration" in the manual to refer to the syntactic form of a declaration (that is, a certain subset of the commands). What about "name" here?

Kha: I don't see a better-fitting word than "definition" here.

PR #628, opened 2025-10-24
fix: broken example

david-christiansen:
The recursive instance example secretly relied on a block with an error in it.

This should un-break CI more generally.


david-christiansen:
CC @jcreedcmu

PR #632, opened 2025-10-24
fix: incorrect attribute name

### LINKED ISSUE #446: `induction_elim` should read `cases_elim` at one point
> The induction_eliminator attribute registers an eliminator for use by the [induction](https://lean-lang.org/doc/reference/latest//Tactic-Proofs/Tactic-Reference/#induction) tactic.
>
>
> attr ::= ...
>     | induction_eliminator
>
> The induction_eliminator attribute registers an eliminator for use by the [cases](https://lean-lang.org/doc/reference/latest//Tactic-Proofs/Tactic-Reference/#cases) tactic.

> attr ::= ...
>    | cases_eliminator

In what way is the content incorrect? 

The second paragraph should use the right attribute name.

david-christiansen:
Closes #446

### DIFF
--- a/Manual/Tactics/Reference.lean
+++ b/Manual/Tactics/Reference.lean
@@ -428,7 +428,7 @@ The {attr}`induction_eliminator` attribute registers an eliminator for use by th
 induction_eliminator
 ```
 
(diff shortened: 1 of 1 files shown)


david-christiansen:
Let's see if the queue does the right thing here

PR #644, opened 2025-10-30
chore: bump max heartbeats in several files

robsimmons:
The changes in leanprover/verso#603 result in more work being done all at once at the end of a document, rather than progressively per-block as the document is read and elaborated. That change pushes a couple of code-heavy manual files over the default heartbeat limit; this PR updates the heartbeat limit in those files.

### DIFF
--- a/Manual/BasicTypes/List.lean
+++ b/Manual/BasicTypes/List.lean
@@ -20,6 +20,7 @@ open Verso.Genre Manual
 open Verso.Genre.Manual.InlineLean
 
 set_option pp.rawOnError true
+set_option maxHeartbeats 1000000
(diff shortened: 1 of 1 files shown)


Kha:
You might instead want to use `Lean.Core.withCurrHeartbeats` to make sure heartbeats are still scoped to sensible work item sizes; anything that scales to the length of the file likely isn't a good fit for heartbeat limiting

robsimmons:
That's some high-quality esoteric knowledge right there @Kha ! Implemented in leanprover/verso#614 , definitely seems to fix the issue more robustly so I'll close this issue.

PR #647, opened 2025-11-05
chore: tell robots not to index non-latest manual

jcreedcmu:
Use meta tag to tell robots to not index any version of the manual other than https://lean-lang.org/doc/reference/latest/ . This includes not indexing the version of the manual that is the current referent of `latest`, e.g., at time of writing, https://lean-lang.org/doc/reference/4.25.0-rc2/ would instruct crawlers not to index.


deploy/overlay.py
```diff
@@ -64,7 +95,7 @@ def deploy_overlays(deploy_dir, src_branch, tgt_branch):
         # (see https://stackoverflow.com/questions/4911794/git-command-for-making-one-branch-like-another/4912267#4912267 for context)
         run_git_command(["git", "merge", "-m", "merge overlays", "--no-ff", "--no-edit", "-s",  "ours", tgt_branch])
         run_git_command(["git", "switch", tgt_branch])
-        run_git_command(["git", "reset", "--hard", src_branch])
+        run_git_command(["git", "reset", "--hard", src_branch, "--"])
```
robsimmons, line 98: What does the extra `--` do?

jcreedcmu: In cases where there is a branch named `foo` and also a file named `foo`, it disambiguates that I'm talking about the branch, not the file. If you wanted to name the file, you'd put the `--` before `src_branch` rather than after. This actually came up because `deploy` is a branch and also a directory that exists in the content of branch `main` right now.

robsimmons (approved):

PR #650, opened 2025-11-12
chore: enable module system

david-christiansen:
This should fix issues with building PRs against nightlies.


david-christiansen:
@robsimmons This should fix the issue with the Lean PR CI, once this change also makes it into nightlies.

PR #659, opened 2025-11-17
chore: bump to nightly-2025-11-14

kim-em:
I have just switched all references to `String.Iterator` to `String.Legacy.Iterator`, but made no attempt to document why this is the legacy version, or what is now preferred.


david-christiansen:
I took care of this separately - but thanks!

PR #660, opened 2025-11-17
fix: restore working error explanation feedback and fix misc warnings

robsimmons:
All examples of the feedback that Lean gives on broken code in error explanations has been busted for about two months, [see here for example](https://lean-lang.org/doc/reference/4.25.0/Error-Explanations/lean___dependsOnNoncomputable/#lean___dependsOnNoncomputable):

image

My hypothesis is something changed about how context objects carry/clear error messages, after some investigation this seems like a mostly minimal diff that mostly addresses the problem.

PR also fixes some long-line warnings while it's at it


robsimmons:
Well huh. @david-christiansen I'm a bit flummoxed, this more-or-less works on my machine:

image

but does not work at all in the preview branch this PR generated:

image

PR #663, opened 2025-11-18
feat: walk-through of type class resolution

robsimmons:
Gives an example for the `#synth` interaction, but also gives a bit of a walk-through of the existing type class resolution tracing mechanism inspired by the discussion from Gray @willcrichton and Krishnamurthi's [PLDI paper](https://maint.dlnext.acm.org/).

I think the actual tracing view could be improved with some observations from their work, but a bit of explanatory text will go some distance.

The two additions are an example of using `#synth` in [the overview of interacting with lean](https://pr-663--lean-reference-manual-review.netlify.app/Interacting-with-Lean/#hash-synth) and a walk-through of synth instance tracing in [Instance Search Problems](https://pr-663--lean-reference-manual-review.netlify.app/Type-Classes/Instance-Synthesis/#instance-search).

### DIFF
--- a/Manual/Classes/InstanceSynth.lean
+++ b/Manual/Classes/InstanceSynth.lean
@@ -90,6 +90,7 @@ This trace can be used to understand how instance synthesis succeeds and why it 
 
 :::paragraph
 Here, we can see the steps Lean takes to conclude that there exists an element of the type {lean}`(Nat ⊕ Empty)` (specifically the element {lean}`Sum.inl 0`):
+Clicking a "▶" symbol expands that branch of the trace, and clicking the "▼" collapses an expanded branch.
 
 ```lean (name := trace)
 set_option pp.explicit true in
@@ -99,7 +100,7 @@ set_option trace.Meta.synthInstance true in
 ```comment
 IF THE LEAN OUTPUT BELOW CHANGES, IT MAY ALSO BE NECESSARY TO UPDATE THE NARRATIVE VERSION OF THIS STORY THAT FOLLOWS
 ```
-```leanOutput trace (expandTrace := Meta.synthInstance)
+```leanOutput trace (expandTrace := Meta.synthInstance) (expandTrace := Meta.synthInstance.resume)
 [Meta.synthInstance] ✅️ Nonempty (Sum Nat Empty)
   [Meta.synthInstance] new goal Nonempty (Sum Nat Empty)
     [Meta.synthInstance.instances] #[@instNonemptyOfInhabited, @instNonemptyOfMonad, @Sum.nonemptyLeft, @Sum.nonemptyRight]
@@ -135,17 +136,19 @@ IF THE LEAN OUTPUT BELOW CHANGES, IT MAY ALSO BE NECESSARY TO UPDATE THE NARRATI
     [Meta.synthInstance.tryResolve] ✅️ Inhabited Nat ≟ Inhabited Nat
     [Meta.synthInstance.answer] ✅️ Inhabited Nat
   [Meta.synthInstance.resume] propagating Inhabited Nat to subgoal Inhabited Nat of Nonempty Nat
+    [Meta.synthInstance.resume] size: 1
+    [Meta.synthInstance.answer] ✅️ Nonempty Nat
   [Meta.synthInstance.resume] propagating Nonempty Nat to subgoal Nonempty Nat of Nonempty (Sum Nat Empty)
+    [Meta.synthInstance.resume] size: 2
+    [Meta.synthInstance.answer] ✅️ Nonempty (Sum Nat Empty)
   [Meta.synthInstance] result @Sum.nonemptyLeft Nat Empty (@instNonemptyOfInhabited Nat instInhabitedNat)
 ```
 :::
 
 :::paragraph
-The Lean InfoView and the live lean editor (available via the "Live ↪" button below) show a clickable "▶" symbol in traces.
-Clicking the "▶" symbol expands that branch of the trace.
-By expanding these arrows, it is possible to retrace the depth-first, backtracking search that Lean uses for type class instance search.
+By exploring the trace, it is possible to follow the depth-first, backtracking search that Lean uses for type class instance search.
 This can take a little practice to get used to!
-For the example above, Lean follows these steps:
+In the example above, Lean follows these steps:
 
 * Lean considers the first goal, {lean}`Nonempty (Sum Nat Empty)`. Lean sees four ways of possibly satisfying this goal:
   - The {name}`Sum.nonemptyRight` instance, which would create a sub-goal {lean}`Nonempty Empty`.
@@ -166,7 +169,7 @@ For the example above, Lean follows these steps:
 :::
 
 The third and fourth original sub-goals are never considered.
-Once the search for {lean}`Nonempty Nat` succeeds, the {keywordOf Lean.Parser.Command.synth}`#synth` outputs the solution:
+Once the search for {lean}`Nonempty Nat` succeeds, the {keywordOf Lean.Parser.Command.synth}`#synth` command finishes and outputs the solution:
 ```leanOutput trace
 @Sum.nonemptyLeft Nat Empty (@instNonemptyOfInhabited Nat instInhabitedNat)
 ```


Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
 Stuck searches may be re-invoked when the elaborator has discovered one of the previously-unknown implicit arguments.
 If this does not occur, stuck searches become failures.
 
+:::example "Tracing instance search"
+
+The {option}`trace.Meta.synthInstance` option provides a powerful tool for
```
david-christiansen: This sentence trails off. What about:

```suggestion
Setting the {option}`trace.Meta.synthInstance` option to {lean}`true` causes Lean to emit an interactive trace of instance synthesis.
This trace can be used to optimize instance synthesis and to understand failures. 
```

david-christiansen: Then, I think the next code block and message should be introduced with a little explanatory sentence, rather than sitting there without an explicit connection to the text. Something like:

```
:::paragraph
Here, Lean synthesizes an instance of {inst}`Nonempty (Nat ⊕ Empty)` by first synthesizing an instance of {inst}`Nonempty Nat` via the {inst}`Inhabited Nat` instance:
```

And then configure the output block to expand the trace by default, so readers can just follow the text.

Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
...
+```lean (name := trace)
+set_option pp.explicit true in
+set_option trace.Meta.synthInstance true in
+#synth Nonempty (Nat ⊕ Empty)
+```
+```leanOutput trace
+[Meta.synthInstance] ✅️ Nonempty (Sum Nat Empty)
+```
+
+The online version of the manual, the Lean InfoView, and the live lean editor (available via the "Live ↪" button below) show a clickable "▶" symbol which will iteratively allow a more thorough investigation of how Lean succeeds, or fails, at type class instance synthesis.
```
david-christiansen: Nitpick: I'm not sure that the adverb "iteratively" meaningfully applies to "allow" here. What about:
```suggestion
The online version of this manual, the Lean InfoView, and the live lean editor (available via the "Live ↪" button below) include a clickable "▶" symbol in the traces.
Clicking it expands that branch of the trace.
This allows an iterative and thorough investigation of how Lean succeeds, or fails, at type class instance synthesis.
```

david-christiansen: That also goes for shorter sentences, which are (all else equal) helpful when readers are feeling cognitively burdened by learning already.

david-christiansen: If we're going to refer to the specific interface here, then we should also have some kind of embedded test that fails when this interface changes, or alternatively abstract `Live ↪` to a string constant that we include in both the UI and the text. I could see this changing, and us forgetting to update this text.

Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
...
+By expanding these arrows, it is possible to retrace the depth-first, backtracking search that Lean uses for type class instance search.
+This can take a little practice to get used to!
+For the example above, the story begins like this:
```
david-christiansen:
```suggestion
For the example above, Lean follows these steps:
```

This description is fragile and should be tested so we can update it when instances are moved around.
A fully-expanded version of the trace should be included in this code here, even if set to invisible, so that if the process changes we'll get a build failure and will know to adapt the description (there should also be a comment to this effect so eager bumpers of versions don't just click "accept changes" without fixing the text).

Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
...
+
+* Lean considers the first goal, {lean}`Nonempty (Sum Nat Empty)`. Lean sees four ways of possibly satisfying this goal:
+  - The {name}`Sum.nonemptyRight` instance, which would create a sub-goal {lean}`Nonempty Empty`.
+  - The {name}`Sum.nonemptyLeft` instance, which would create a sub-goal {lean}`Nonempty Nat`.
+  - The {name}`instNonemptyOfMonad` instance, which would create two sub-goals {lean}`Monad (Sum Nat)` and {lean}`Nonempty Nat`.
+  - The {name}`instNonemptyOfInhabited` instance, which would create a sub-goal {lean}`Inhabited (Sum Nat Empty)`.
+* The first sub-goal, {lean}`Nonempty Empty`, is considered. Lean sees two ways of possibly satisfying this goal:
+  - The {name}`instNonemptyOfMonad` instance, though this is a non-starter because {lean}`Empty` does not have the structure of a monadic type, which must be a type constructor applied to another type. Lean describes this as a failure of {option}`trace.Meta.synthInstance.tryResolve` to solve the equation `Nonempty Empty ≟ Nonempty (?m.5 ?m.6)`.
```
david-christiansen: This sentence is complicated! What about:
```suggestion
  - The {name}`instNonemptyOfMonad` instance is rejected.
      It can't be used because {lean}`Empty` does not have the structure of a monadic type, which must be an application of the monad to a type.
      Lean describes this as a failure of {option}`trace.Meta.synthInstance.tryResolve` to solve the equation `Nonempty Empty ≟ Nonempty (?m.5 ?m.6)`, where `?m.5` is the monad.
```
This also removes a subtle incorrectness: the monad doesn't need to be a type constructor in the sense of "canonical type former", because we can give monad instances for constants defined with e.g. `def` too.

Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
...
+  - The {name}`instNonemptyOfInhabited` instance, which would create a sub-goal {lean}`Inhabited Empty`.
+* The newly-generated sub-goal, {lean}`Inhabited Empty`, is considered. Lean only sees one way of possibly satisfying this goal, {name}`instInhabitedOfMonad`, which is a non-starter because {lean}`Empty` does not have the structure of a monadic type.
```
david-christiansen:
```suggestion
* The newly-generated sub-goal, {lean}`Inhabited Empty`, is considered.
   Lean only sees one way of possibly satisfying this goal, {name}`instInhabitedOfMonad`; this can't be used because {lean}`Empty` is not a monad applied to a type.
```
I think this is a bit more specific.

Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
...
+* Backtracking to the second original sub-goal, {lean}`Nonempty Nat`. This sub-goal eventually succeeds.
```
david-christiansen:
```suggestion
* The search backtracks to the second original sub-goal, {lean}`Nonempty Nat`.
   This search eventually succeeds.
``` 
Because a goal can't succeed. An agent can achieve their goal, and thus have a success, however.

Manual/Interaction.lean
```diff
@@ -269,10 +269,45 @@ tag := "hash-synth"
 ```grammar
 #synth $t
 ```
+:::
 
-The {keywordOf Lean.Parser.Command.synth}`#synth` command attempts to synthesize an instance for the provided class.
+The {keywordOf Lean.Parser.Command.synth}`#synth` command invokes Lean's {tech}[type class] resolution machinery and attempts to perform {ref "instance-synth"}[instance synthesis] to find an instance for the given type class.
 If it succeeds, then the resulting instance term is output.
 
+:::example "Synthesizing a type class instance"
```
david-christiansen: Example titles should use Chicago title case

Manual/Classes/InstanceSynth.lean
```diff
@@ -81,6 +83,41 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
 Stuck searches may be re-invoked when the elaborator has discovered one of the previously-unknown implicit arguments.
 If this does not occur, stuck searches become failures.
 
+:::example "Tracing instance search"
```
david-christiansen: Example titles should use Chicago title case

Manual/Classes/InstanceSynth.lean
david-christiansen: We should also have a proper reference entry for this option. Can you add one as part of this, ideally next to these examples, so the search feature finds it directly?

Manual/Interaction.lean
```diff
@@ -269,10 +269,45 @@ tag := "hash-synth"
...
+
+This example shows that Lean will let us add two integers, and the result will be an integer:
```
david-christiansen, line 282: I think it's always good to be explicit about connections:
```suggestion
The `+` operator is notation for a call to {name}`HAdd.hAdd`, which is the single method in the {name}`HAdd` type class.
The existence of an instance of {inst}`HAdd Int Int Int` indicates that adding two integers returns an integer:
```
There should also be `:::paragraph` around this and the code.

Manual/Interaction.lean
```diff
@@ -269,10 +269,45 @@ tag := "hash-synth"
...
+
+```lean (name := synthInstHAddNat)
+#synth HAdd Int Int Int
+```
+```leanOutput synthInstHAddNat
+instHAdd
+```
+By default the output term does not show its arguments.
+Setting the {option}`pp.explicit` option lets us see more:
```
david-christiansen:
```suggestion
Setting the option {option}`pp.explicit` to `true` causes Lean to display them:
```

Manual/Interaction.lean
```diff
@@ -269,10 +269,45 @@ tag := "hash-synth"
...
+
+This example shows that Lean will let us add two integers, and the result will be an integer:
+
+```lean (name := synthInstHAddNat)
+#synth HAdd Int Int Int
+```
+```leanOutput synthInstHAddNat
+instHAdd
+```
+By default the output term does not show its arguments.
```
david-christiansen:
```suggestion
By default, Lean does not show the implicit arguments in the output term.
Instance arguments are implicit, however, which decreases the usefulness of this output for understanding instance synthesis.
```

Manual/Interaction.lean
david-christiansen: Or together with preceding suggestion:
```suggestion
Setting the option {option}`pp.explicit` to `true` causes Lean to display the implicit arguments, including instances:
```

david-christiansen: This also needs `:::paragraph` because the code and text are part of a semantic paragraph

Manual/Classes/InstanceSynth.lean
david-christiansen: This paragraph and the list are really the same paragraph and should be indicated as such using `:::paragraph`

Manual/Interaction.lean
```diff
@@ -269,10 +269,45 @@ tag := "hash-synth"
...
+
+```lean (name := synthInstHAddNatInt) +error
+#synth HAdd Int String String
+```
+```leanOutput synthInstHAddNatInt
+failed to synthesize
+  HAdd Int String String
+
+Hint: Additional diagnostic information may be available using the `set_option diagnostics true` command.
+```
```
david-christiansen, line 314:
````suggestion
:::paragraph
Lean does not allow the addition of integers and strings, as demonstrated by this failure of type class instance synthesis:

```lean (name := synthInstHAddNatInt) +error
#synth HAdd Int String String
```
```leanOutput synthInstHAddNatInt
failed to synthesize
  HAdd Int String String

Hint: Additional diagnostic information may be available using the `set_option diagnostics true` command.
```
:::
````

File: Manual/Classes/InstanceSynth.lean [hunk 13]
```diff
@@ -81,6 +83,94 @@ Search may succeed, fail, or get stuck; a stuck search may occur when an unknown
...
+
+```lean (name := trace)
+set_option pp.explicit true in
+set_option trace.Meta.synthInstance true in
+#synth Nonempty (Nat ⊕ Empty)
+```
+```comment
+IF THE LEAN OUTPUT BELOW CHANGES, IT MAY ALSO BE NECESSARY TO UPDATE THE NARRATIVE VERSION OF THIS STORY THAT FOLLOWS
+```
+```leanOutput trace (expandTrace := Meta.synthInstance)
```
robsimmons: @david-christiansen does this seem a viable way to signal the fragile dependency on lean's output?

PR #670, opened 2025-11-25
feat: integrate module docs with other docs

david-christiansen:
This PR splits up and expands the module system chapter, integrating it with the relevant descriptions of other features. Modules are on their way to being a core feature of Lean, and should be documented as such.


david-christiansen:
It seems the live links already do the right thing here, due to the choice of underlying blocks.

PR #672, opened 2025-11-25
feat: start work on iterators and ranges

david-christiansen:


Manual/Iterators.lean
```diff
@@ -0,0 +1,1123 @@
...
+Additionally, other built-in types such as ranges support iteration using the same convention.
+:::
+
+# Run-Time Considerations
+
+For many use cases, using iterators can give a performance benefit by avoiding allocating intermediate data structures.
+Without iterators, zipping a list with an array requires first converting one of them to the other type, allocating an intermediate structure, and then using the appropriate {name List.zip}`zip` function.
+Using iterators, the intermediate structure can be avoided.
+
+When an iterator is consumed, the resulting computation should be though of as a single loop, even if the iterator itself is build using combinators from a number of underlying iterators.
```
datokrat:
```suggestion
When an iterator is consumed, the resulting computation should be thought of as a single loop, even if the iterator itself is built using combinators from a number of underlying iterators.
```

Manual/Iterators.lean
```diff
@@ -0,0 +1,1123 @@
...
+One step of the loop may carry out multiple steps from the underlying iterators.
+In many cases, the Lean compiler can optimize iterator computations, removing the intermediate overhead, but this is not guaranteed.
+When profiling shows that significant time is taken by a tight loop that involves multiple sources of data, it can be necessary to inspect the compiler's IR to see whether the iterators' operations were fused.
+In particular, if the IR contains many pattern matches over steps, then it can be a sign of a failure to inline or specialize.
+In other cases, it may be necessary to write a tail-recursive function by hand rather than using the higher-level API.
```
datokrat: Given that specialization failure is a sign that a tail-recursive function would be more efficient, I find the formulation "in other cases" misleading. How about:

```suggestion
In particular, if the IR contains many pattern matches over steps, then it can be a sign of a failure to inline or specialize.
If this is the case, it may be necessary to write a tail-recursive function by hand rather than using the higher-level API.
```

Manual/Iterators.lean
```diff
@@ -0,0 +1,1123 @@
...
+{docstring PlausibleIterStep.done}
+
+## Finite and Productive Iterators
+
+:::paragraph
+Not all iterators are guaranteed to return a finite number of results; it is perfectly sensible to iterate over all of the natural numbers.
+Similarly, not all iterators are guaranteed to either return a single result or terminate; iterators may be defined using arbitrary programs.
+Thus, Lean divides iterators into three termination classes:
+* {deftech (key:="finite iterator")}_Finite_ iterators are guaranteed to finish iterating after a finite number of steps. These iterators have a {name}`Finite` instance.
+* {deftech (key:="productive iterator")}_Productive_ iterators are guaranteed to yield a value or terminate in finite time, but they may yield infinitely many values. These iterators have a {name}`Productive` instance.
```
datokrat: Let me just note on a little subtlety that I'm struggling with when talking about termination in Lean. By "terminate in finite time", we clearly mean that the iterator is done after finitely many steps. However, there's no a priori guarantee that the step function itself terminates. So it's theoretically possible that a `Finite` iterator's loop will run forever if the step function is sufficiently insidious.

I think that mentioning this would be more confusing than helpful and it's not relevant in 99% of cases, so from a pragmatic view on language, I think the explanation is great as you wrote it.

Manual/Iterators.lean
```diff
@@ -0,0 +1,1123 @@
...
+
+{docstring Iter.finitelyManySkips}
+
+{docstring IterM.finitelyManySkips}
+
+{docstring IterM.TerminationMeasures.Productive +allowMissing}
+
+{docstring Iter.attachWith}
+
+{docstring IterM.attachWith}
```
datokrat: `attachWith` doesn't really help with termination proofs. Perhaps these are better placed among the combinators?

Manual/Iterators.lean
```diff
@@ -0,0 +1,1123 @@
...
+it.double    ---a-a---b-b---c-c---d-d⊥
+```
+:::
+:::paragraph
+A marble diagram for an iterator combinator that duplicates each element of the underlying iterator looks like this:
+```
+it           ---a  ---b  ---c  ---d⊥
+it.double    ---a-a---b-b---c-c---d-d⊥
+```
+:::
```
datokrat: If I'm not mistaken, this example is a duplicate.

```suggestion
```

Manual/Iterators.lean
```diff
@@ -0,0 +1,1123 @@
...
+
+{docstring Std.Iterators.IterM.filterMapM}
+
+{docstring Std.Iterators.IterM.filterMapWithPostcondition}
+
+{docstring Std.Iterators.IterM.zip}
+
+# Reasoning About Iterators
+
+## Equivalence
```
datokrat: I would bring the `Equivalence` section last because the other reasoning techniques are usually more helpful. I propose the following order:

* Induction Principles
* Monads for Reasoning
* Equivalence

However, probably the most useful way to reason about iterators is to use the lemmas provided in `Init.Data.Iterators.Lemmas`. While it would not be helpful to state the gazillions of lemmas in the reference manual, I deem it very helpful to mention that these are available, lest people try doing everything by hand. Something like the following outline seems reasonable to me:

* Reasoning about Consumers: When reasoning about high-level consumers such as `toList`, it is easiest to use the lemmas that are available. I think it is sufficient to list very few examples such as `List.toList_iter` and `Iter.toList_map` or provide a simple example where these lemmas are being applied.
* Stepwise Reasoning: This is the place where the induction principles have their place. I think it is also worth mentioning that there are lemmas for the stepwise behavior of all the producers/combinators because they are necessary to complete the inductive proofs.
* Monads for Reasoning
* Equivalence

datokrat: A simple lemma that might just so be nontrivial enough to be interesting could be this one:

```lean
example (l : List Nat) :
    (l.iter.map (· * 2)).all (· % 2 = 0) := by
  simp [← Std.Iterators.Iter.all_toList] -- simp lemmas are at work here
```

It illustrates that for *pure* iterators, most statements can be proved comfortably by reducing the statements into statements about lists. In this case, this is done by pulling `toList`/`toArray` to the inside. For monadic iterators, one will often need to either reason stepwise or use `mvcgen`.

Manual/Iterators.lean
david-christiansen: I actually think that "finitely many steps" is a better phrasing. I'll go update it.

Manual/Iterators.lean
david-christiansen: Ah, I'd thought of the way that `attach` is useful to help wf-rec go through in nested inductive types. I'll go move them.

Manual/Iterators.lean
david-christiansen: I've been looking for a way to get the gazillions of lemmas into the manual in a genuinely useful way. I've got a half-baked design in my head (at least for browsers with JavaScript) that I hope to get to at some point :)

Manual/Iterators.lean
david-christiansen: Done

Manual/Iterators.lean
david-christiansen: Thanks! Fixed.

Manual/Iterators.lean
david-christiansen: OK, I think I've updated it as described here. Do you mind doing a quick once-over to see if I missed anything important? Thanks!

datokrat: Looks good to me, thanks again.

datokrat (approved):

PR #683, opened 2025-12-04
feat: mvcgen reference

david-christiansen:


Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
-theorem mySum_correct_vanilla (l : Array Nat) : mySum l = l.sum := by
-  -- Turn the array into a list
-  cases l with | mk l =>
-  -- Unfold `mySum` and rewrite `forIn` to `foldl`
-  simp [mySum]
-  -- Generalize the inductive hypothesis
-  suffices h : ∀ out, List.foldl (· + ·) out l = out + l.sum by simp [h]
-  -- Grind away
-  induction l with grind
+def ItIsSecret : SPred [String] := (⟨· = "secret"⟩)
```
sgraf812: I suppose `⟨⟩` to inject into `ULift` works, but maybe it would be better to use corner brackets here so that we use `SPred.pure` instead of the defeq `SPred [] = ULift Prop`? Haven't tested whether that'll work with cdot though. In case it doesn't work, I'm happy with what you have.

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
 
-Exceptions are the reason why the type of postconditions {lean}`PostCond α ps` is not simply a single condition of type {lean}`α → Assertion ps` for the success case.
-To see why, suppose the latter was the case, and suppose that program {lean}`prog` throws an exception in a prestate satisfying {lean}`P`.
-Should we be able to prove {lean}`⦃P⦄ prog ⦃⇓ r => Q' r⦄`?
-(Recall that `⇓` is grammatically similar to `fun`.)
-There is no result `r`, so it is unclear what this proof means for {lean}`Q'`!
+{docstring StateM.of_wp_run'_eq}
 
-So there are two reasonable options, inspired by non-termination in traditional program logics:
+{docstring StateM.of_wp_run'_eq}
```
sgraf812:
```suggestion
```

duplicate line

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
+:::leanSection
+```lean -show
+variable [WP m ps] {x : m α} {P : Assertion ps} {Q : PostCond α ps}
+```
+{lean}`⦃P⦄ x ⦃Q⦄` is syntactic sugar for {lean}`Triple x P Q`.
+:::
+::::
 
-{includeDocstring Std.Do.PostCond.noThrow}
+{docstring SVal}
```
sgraf812: This seems a bit out of place here. Move to `SPred`?

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
+{deftech}_Specification lemmas_ are designated theorems that associate Hoare triples with functions.
+When {tactic}`mvcgen` encounters a function, it checks whether there are any registered specification lemmas and attempts to use them to discharge intermediate {tech}[verification conditions].
+If there is no applicable specification lemma, then the connection between the statement's pre- and postconditions will become a verification condition.
+Specification lemmas allow compositional reasoning about libraries of monadic code.
 
-```lean -show
-namespace Exceptions
+When applied to a theorem whose statement is a Hoare triple, the {attr}`spec` attribute registers the theorem as a specification lemma.
+These lemmas are used in order of priority.
+For theorems, the {keywordOf simpPre}`↑`, {keywordOf simpPost}`↑`, and {keyword}`←` specifiers are ignored.
```
sgraf812:
```suggestion
For theorems, the {keywordOf simpPre}`↓`, {keywordOf simpPost}`↑`, and {keyword}`←` specifiers are ignored.
```

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
+
+The {attr}`spec` attribute may also be applied to definitions.
+On definitions, it indicates that the definition should be simplified during verification condition generation.
+For definitions, {attr}`spec` uses the {keywordOf simpPre}`↑`, {keywordOf simpPost}`↑`, and {keyword}`←` specifiers in the same manner as {tactic}`simp`.
```
sgraf812:
```suggestion
For definitions, {attr}`spec` uses the {keywordOf simpPre}`↓`, {keywordOf simpPost}`↑`, and {keyword}`←` specifiers in the same manner as {tactic}`simp`.
```

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
+Specification lemmas allow compositional reasoning about libraries of monadic code.
 
-```lean -show
-namespace Exceptions
+When applied to a theorem whose statement is a Hoare triple, the {attr}`spec` attribute registers the theorem as a specification lemma.
+These lemmas are used in order of priority.
+For theorems, the {keywordOf simpPre}`↑`, {keywordOf simpPost}`↑`, and {keyword}`←` specifiers are ignored.
+
+The {attr}`spec` attribute may also be applied to definitions.
+On definitions, it indicates that the definition should be simplified during verification condition generation.
```
sgraf812:
```suggestion
On definitions, it indicates that the definition should be unfolded during verification condition generation.
```

Manual/VCGen/Tutorial.lean
```diff
@@ -0,0 +1,975 @@
...
+for some {lean}`p`.
+Pure, stateful hypotheses may be freely moved into the regular Lean context and back.
+(This can be done manually with the {tactic}`mpure` tactic.)
+
+:::::
+
+## Composing Specifications
+
+Nested unfolding of definitions as in {tactic}`mvcgen`{lit}` [`{name}`mkFreshN`{lit}`, `{name}`mkFresh`{lit}`]` is quite blunt but effective for small programs.
+A more compositional way is to develop individual {deftech}_specification lemmas_ for each monadic function.
```
sgraf812:
```suggestion
A more compositional way is to develop individual {tech}_specification lemmas_ for each monadic function.
```

sgraf812: I think there might be a few `deftech`s in the tutorial that should be `tech`s now.

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
 ```
+It is correct if it is equal to {name}`List.reverse`.
+However, {tactic}`mvcgen` does not make the goal easier to prove:
+```lean +error -keep (name := noInst)
+theorem rev_correct :
+    (rev xs).run = xs.reverse := by
+  generalize h : (rev xs).run = x
+  apply Identity.of_wp_run_eq h
+  simp only [rev]
+  mvcgen
```
sgraf812: Could be just

```suggestion
  mvcgen [rev]
```

depending on whether you think that is advisable at this point.

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
-instance Result.instLawfulMonad : LawfulMonad Result := by
-  -- TODO: Replace sorry with grind when it no longer introduces section
-  --       variables
-  apply LawfulMonad.mk' <;> (simp only [Result.instMonad]; sorry)
+variable [Monad m] [WPMonad m ps] {e : m α} {P : Assertion ps} {Q : PostCond α ps}
+```
+The verification conditions for a goal are generated as follows:
+1. A number of simplifications and rewrites are applied.
+2. The goal should now be of the form {lean}`P ⊢ₛ wp⟦e⟧ Q` (that is, an entailment from some set of stateful assumptions to the weakest precondition that implies a desired postcondition).
+3. {tech}[Reducible] constants in the expression {lean}`e` are unfolded, and definitions marked {attrs}`@[spec]` are simplified.
```
sgraf812:
```suggestion
3. {tech}[Reducible] constants and definitions marked {attrs}`@[spec]` in the expression {lean}`e` are unfolded.
```

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
-To successfully discharge such a VC, `mvcgen` comes with an entire proof mode that is inspired by that of the Iris concurrent separation logic.
-(In fact, the proof mode was adapted in large part from its Lean clone, [`iris-lean`](https://github.com/leanprover-community/iris-lean).)
-The {ref "tactic-ref-spred"}[tactic reference] contains a list of all proof mode tactics.
+This specification lemma for {name}`bump` is proved in an intentionally low-level manner to demonstrate the intermediate proof states:
+```lean
+theorem bump_correct :
+      ⦃ fun n => ⌜n = k⌝ ⦄
+      bump (m := m) i
+      ⦃ ⇓ r n => ⌜r = n ∧ n = k + i⌝ ⦄ := by
+  mvcgen
```
sgraf812: Since `mvcgen` gives up almost immediately, it is equivalent to just
```suggestion
  mintro hnk
```

Manual/VCGen.lean
```diff
@@ -28,1004 +29,884 @@ set_option mvcgen.warning false
...
+  . simp_all
+```
 
+The lemma can also be proved using only the simplifier:
+```lean
+theorem bump_correct' :
+    ⦃ fun n => ⌜n = k⌝ ⦄
+    bump (m := m) i
+    ⦃ ⇓ r n => ⌜r = n ∧ n = k + i⌝ ⦄ := by
+  mvcgen
```
sgraf812: Dito

```suggestion
  mintro _
```

sgraf812 (dismissed):
Really nice, thank you! We may want to delete some redundant info from the tutorial in the future, but that can wait. Just some small nits

Manual/VCGen.lean
sgraf812: On second thought, you perhaps did this to work around something related to https://github.com/leanprover/lean4/issues/11509?

Manual/VCGen.lean
david-christiansen: It did work with cdot, but it was actually nicer to use an explicit lambda.

Manual/VCGen.lean
david-christiansen: Done!

Manual/VCGen/Tutorial.lean
david-christiansen: I went through and checked them all. Good thought.

Manual/VCGen.lean
david-christiansen: No, it was just a bigger hammer than needed for that nail.

Manual/VCGen.lean
david-christiansen: Done.

david-christiansen:
I'll take your Zulip comment as an approval, despite the "request changes" flag being set here. Thanks!

PR #684, opened 2025-12-05
feat: document well-founded reduction behaviour

### LINKED ISSUE #680: Document new wfrec reduction behavior for Nat-valued measures
As soon as nightly-testing is up-to-date with 2025-12-02, which includes https://github.com/leanprover/lean4/pull/7965, I should document the new behavior.

nomeata:
This PR documents well-founded reduction behaviour. This fixes #680.

### DIFF
--- a/Manual/RecursiveDefs/WF.lean
+++ b/Manual/RecursiveDefs/WF.lean
@@ -717,6 +717,54 @@ Try this:
 
 :::
 
+# Reduction behaviour
+%%%
+tag := "well-founded-reduction"
+%%%
+
+By default, functions defined by well-founded recursion are marked as {attr}`irreducible`, which
+means that the actual internal construction (described {ref "well-founded-theory"}[below]) is
+not visible.
+
+This means that proofs using tactics like {tactic}`rfl` or {tactic}`decide` will not be able to
+unfold such functions. Instead, proofs that use rewriting ({tactic}`unfold`, {tactic}`rw`,
+{tactic}`simp`) need to be used to reason about functions defined by well-founded recursion.
+
+Trying to override this using an explicit {attr}`semireducible` annotation will produce a warning:
+
+```lean (name := semireducibleack)
+@[semireducible] def ack : Nat → Nat → Nat
+  | 0,     n     => n + 1
+  | m + 1, 0     => ack m 1
+  | m + 1, n + 1 => ack m (ack (m + 1) n)
+termination_by m n => (m, n)
+```
+
+```leanOutput semireducibleack
+marking functions defined by well-founded recursion as `semireducible` is not effective
+```
+
+This is ineffective as reducing the internal construction of well-founded recursion would require reducing the well-foundedness proof, which is often not possible (for example because they are not available due to the module system) or dangerously expensive.
+To avoid these issues, the internal construction prevents reduction of the proof.
+
+In the special case where the termination {tech}[measure] is a single value of type {lean}`Nat`, a specialized fix-point operator is used internally, which has reasonable reduction behaviour.
+So while still {attr}`irreducible` by default, such functions can be marked as {attr}`semireducible`, then should reduce whenever the measure reduces well.
+
+For arguments that are not ground values, equations are still not definitional equalities:
+
+```lean (name := natwf)
+@[semireducible] def log2 (n : Nat) : Nat :=
+  if n = 0 then 0 else log2 (n / 2) + 1
+termination_by n
+
+example : log2 64 = 7 := by rfl
+example : log2 64 = 7 := by decide
+
+example : log2 n = if n = 0 then 0 else log2 (n / 2) + 1 := by
+  fail_if_success rfl
+  exact log2.eq_def n
+```
+
 # Preprocessing Function Definitions
 %%%
 tag := "well-founded-preprocessing"
@@ -816,6 +864,9 @@ To see the preprocessed function definition, before and after the removal of {na
 {spliceContents Manual.RecursiveDefs.WF.PreprocessExample}
 
 # Theory and Construction
+%%%
+tag := "well-founded-theory"
+%%%
 
 ```lean -show
 section
@@ -845,6 +896,10 @@ The definition of {name}`WellFounded` builds on the notion of _accessible elemen
 {docstring WellFounded}
 
 {docstring Acc}
+
+In the special case where the measure is {lean}`Nat`, Lean uses a specialized fix-point operator, as discussed in {ref "well-founded-reduction"}[reduction behaviour].
+
+{docstring WellFounded.Nat.fix}
 
 ::: example "Division by Iterated Subtraction: Termination Proof"
 


Manual/RecursiveDefs/WF.lean
```diff
@@ -717,6 +717,54 @@ Try this:
...
 
+# Reduction behaviour
+%%%
+tag := "well-founded-reduction"
+%%%
+
+By default, functions defined by well-founded recursion are marked as {attr}`irreducible`, which
+means that the actual internal construction (described {ref "well-founded-theory"}[below]) is
+not visible.
+
```
david-christiansen, line 728: I think these should be one paragraph:

```suggestion
```

Manual/RecursiveDefs/WF.lean
```diff
@@ -717,6 +717,54 @@ Try this:
...
+  | m + 1, n + 1 => ack m (ack (m + 1) n)
+termination_by m n => (m, n)
+```
+
+```leanOutput semireducibleack
+marking functions defined by well-founded recursion as `semireducible` is not effective
+```
+
+This is ineffective as reducing the internal construction of well-founded recursion would require reducing the well-foundedness proof, which is often not possible (for example because they are not available due to the module system) or dangerously expensive.
+To avoid these issues, the internal construction prevents reduction of the proof.
```
david-christiansen, line 748: These are one logical para:

````suggestion
:::paragraph
Trying to override this using an explicit {attr}`semireducible` annotation will produce a warning:

```lean (name := semireducibleack)
@[semireducible] def ack : Nat → Nat → Nat
  | 0,     n     => n + 1
  | m + 1, 0     => ack m 1
  | m + 1, n + 1 => ack m (ack (m + 1) n)
termination_by m n => (m, n)
```

```leanOutput semireducibleack
marking functions defined by well-founded recursion as `semireducible` is not effective
```

This is ineffective as reducing the internal construction of well-founded recursion would require reducing the well-foundedness proof, which is often not possible (for example because they are not available due to the module system) or dangerously expensive.
To avoid these issues, the internal construction prevents reduction of the proof.
:::
````

File: Manual/RecursiveDefs/WF.lean [hunk 3]
```diff
@@ -717,6 +717,54 @@ Try this:
...
+  | m + 1, 0     => ack m 1
+  | m + 1, n + 1 => ack m (ack (m + 1) n)
+termination_by m n => (m, n)
+```
+
+```leanOutput semireducibleack
+marking functions defined by well-founded recursion as `semireducible` is not effective
+```
+
+This is ineffective as reducing the internal construction of well-founded recursion would require reducing the well-foundedness proof, which is often not possible (for example because they are not available due to the module system) or dangerously expensive.
```
david-christiansen, line 747: A definition of "effective" would be useful here if it's a technical term. If not, can we phrase this a bit differently, perhaps something like "unlikely to have the expected results" or even better "unlikely to result in a definition that can be unfolded in practice"?

And should the text also mention axioms as a source of non-reduction, or is that not a major concern here?

Manual/RecursiveDefs/WF.lean
```diff
@@ -717,6 +717,54 @@ Try this:
...
+To avoid these issues, the internal construction prevents reduction of the proof.
+
+In the special case where the termination {tech}[measure] is a single value of type {lean}`Nat`, a specialized fix-point operator is used internally, which has reasonable reduction behaviour.
```
david-christiansen, line 750:
```suggestion
In the special case where the termination {tech}[measure] is a single value of type {lean}`Nat`, a specialized fixpoint operator is used internally, which has reasonable reduction behaviour.
```

Manual/RecursiveDefs/WF.lean
```diff
@@ -717,6 +717,54 @@ Try this:
...
+So while still {attr}`irreducible` by default, such functions can be marked as {attr}`semireducible`, then should reduce whenever the measure reduces well.
```
david-christiansen, line 751:
```suggestion
So while still {attr}`irreducible` by default, such functions can be marked as {attr}`semireducible`.
They should then reduce whenever the measure reduces well.
```

david-christiansen: I think two sentences is easier to parse here.

Manual/RecursiveDefs/WF.lean
```diff
@@ -717,6 +717,54 @@ Try this:
...
+termination_by n
+
+example : log2 64 = 7 := by rfl
+example : log2 64 = 7 := by decide
+
+example : log2 n = if n = 0 then 0 else log2 (n / 2) + 1 := by
+  fail_if_success rfl
+  exact log2.eq_def n
+```
+
```
david-christiansen, line 767: This should be separated into a named `example` block, which will enable links to live and also be consistent with the structure of the text. It would also be nice to get the error message from `rfl` rendered in the manual to make it more clear what's going on. Maybe:

````suggestion
For arguments that are not ground values, the function's {tech}[equational theorems] are still not {tech (key:="definitional equality")}[definitional equalities].

:::example "Reduction of `Nat` Measures"
This implementation of {name}`log2` has a {name}`Nat` measure and is elaborated to a use of a special fixpoint operator that can reduce when provided with ground arguments:
```lean
@[semireducible] def log2 (n : Nat) : Nat :=
  if n = 0 then 0 else log2 (n / 2) + 1
termination_by n
```
Its reduction behavior holds definitionally when its arguments are ground values:
```lean
example : log2 64 = 7 := by rfl
example : log2 64 = 7 := by decide
```
When the argument is a variable, however, the equation does not hold definitionally:
```lean +error (name := nondef)
example : log2 n = if n = 0 then 0 else log2 (n / 2) + 1 := by
  rfl
```
```leanOutput nondef
ERROR HERE
```
In these cases, the function's equational theorems can be used, just as for any function defined by well-founded recursion. 
```lean
example : log2 n = if n = 0 then 0 else log2 (n / 2) + 1 := by
  exact log2.eq_def n
```
:::
````

File: Manual/RecursiveDefs/WF.lean [hunk 7]
```diff
@@ -846,6 +897,10 @@ The definition of {name}`WellFounded` builds on the notion of _accessible elemen
 
 {docstring Acc}
 
+In the special case where the measure is {lean}`Nat`, Lean uses a specialized fix-point operator, as discussed in {ref "well-founded-reduction"}[reduction behaviour].
```
david-christiansen, line 900: We use US English:
```suggestion
In the special case where the measure is {lean}`Nat`, Lean uses a specialized fix-point operator, as discussed in {ref "well-founded-reduction"}[reduction behavior].
```

Manual/RecursiveDefs/WF.lean
```diff
@@ -717,6 +717,54 @@ Try this:
 
 :::
 
+# Reduction behaviour
```
david-christiansen, line 720: We use Chicago-style headlines and US English:
```suggestion
# Reduction Behavior
```

david-christiansen:
Looks like we both let this one drop. Want me to finish it up and get it merged?

nomeata:
Indeed, good find. Yes please!

PR #689, opened 2025-12-09
fix: update the chapter on iterators to account for extrinsic termination

datokrat:
This PR modifies that iterator chapter to accounf for the changes in https://github.com/leanprover/lean4/pull/11038. The consumer functions no longer require `Finite` instances and may run forever. If termination needs to be guaranteed, `Finite` can be proved afterwards or `it.ensureTermination.toList` and consorts can be used to check finiteness in-place.

### DIFF
--- a/Manual/Iterators.lean
+++ b/Manual/Iterators.lean
@@ -261,11 +261,11 @@ All finite iterators are necessarily productive.
 
 {docstring Productive +allowMissing}
 
-Lean's standard library provides many functions that iterate over an iterator. These consumer functions usually do not
+Lean's standard library provides many functions that consume an iterator. These consumer functions usually do not
 make any assumptions about the underlying iterator. In particular, such functions may run forever for certain iterators.
 
 Sometimes, it is of utmost importance that a function does terminate.
-For these cases, {name}`Iter.allowNontermination` provides variants of consumers that are guaranteed to terminate.
+For these cases, {name}`Iter.ensureTermination` provides variants of consumers that are guaranteed to terminate.
 They usually require proof that the involved iterator is finite.
 
 {docstring Iter.ensureTermination}
@@ -307,7 +307,7 @@ instance [Pure m] : Iterator Nats m Nat where
 
 Whenever an iterator is defined, {name}`IteratorCollect` and {name}`IteratorLoop` instances should be provided.
 They are required for most consumers of iterators such as {name}`Iter.toList` or the `for` loops.
-One can use their default implementations as follows:
+While it is possible to provide specially-optimized implementations for some iterators, most iterators should use the default implementations:
 
 ```lean
 instance [Pure m] [Monad n] : IteratorCollect Nats m n :=
@@ -345,13 +345,13 @@ def Nats.iter : Iter (α := Nats) Nat :=
 :::
 
 :::paragraph
-One can print all natural numbers by running the following function:
+This {name}`IO` action prints all the natural numbers:
 ```lean
 def f : IO Unit := do
   for x in Nats.iter do
     IO.println s!"{x}"
 ```
-This function never terminates, printing all natural numbers in increasing order, one
+A call to `f` never terminates; it continues printing all natural numbers in increasing order, one
 after another.
 :::
 
@@ -371,10 +371,11 @@ This iterator is most useful with combinators such as {name}`Iter.zip`:
 :::
 
 :::paragraph
-In contrast to the previous example, this loop terminates because `xs.iter` is a finite iterator,
-One can make sure that a loop actually terminates by providing a {name}`Finite` instance:
+In contrast to {name}`f`, this loop terminates because `xs.iter` is a finite iterator.
+A {name}`Finite` instance ensures that a loop actually terminates:
 ```lean (name := natfin)
-#check type_of% (Nats.iter.zip ["cat", "dog"].iter).internalState
+example : (Zip Nats Id (ListIterator String) String) Id :=
+  (Nats.iter.zip ["cat", "dog"].iter).internalState
 
 #synth Finite (Zip Nats Id (ListIterator String) String) Id
 ```
@@ -527,8 +528,8 @@ With the {name}`IteratorCollect` instance in place, {name}`Iter.toArray` now wor
 #['a', 'b', 'c']
 ```
 
-In general, `Iter.toArray` might run forever. One can prove that `abc` is finite, and the above example will terminate after finitely many steps, by
-constructing a `Finite (Triple Char) Id` instance.
+In general, `Iter.toArray` might run forever.
+To use `abc` in a context where termination is important, it must be proved finite by constructing a `Finite (Triple Char) Id` instance.
 It's easiest to start at {name}`TriplePos.done` and work backwards toward {name}`TriplePos.fst`, showing that each position in turn has a finite chain of successors:
 
 ```lean
@@ -570,7 +571,7 @@ instance [Pure m] : Finite (TripleIterator α) m where
       cases pos <;> grind
 ```
 
-To enable the iterator in {keywordOf Lean.Parser.Term.doFor}`for` loops, an instance of {name}`IteratorLoop` are needed:
+To enable the iterator in {keywordOf Lean.Parser.Term.doFor}`for` loops, an instance of {name}`IteratorLoop` is needed:
 ```lean
 instance [Monad m] [Monad n] :
     IteratorLoop (TripleIterator α) m n :=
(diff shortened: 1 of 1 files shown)


Manual/Iterators.lean
```diff
@@ -261,12 +261,16 @@ All finite iterators are necessarily productive.
 
 {docstring Productive +allowMissing}
 
-Sometimes, a needed {name}`Finite` instance is not available because an iterator has not yet been proved finite.
-In these cases, {name}`Iter.allowNontermination` can be used to bypass a finiteness requirement.
+Lean's standard library provides many functions that iterate over an iterator. These consumer functions usually do not
```
david-christiansen: I think we've mostly settled on the term "consume". Is that right? If so:
```suggestion
Lean's standard library provides many functions that consume an iterator. These consumer functions usually do not
```

Manual/Iterators.lean
```diff
@@ -261,12 +261,16 @@ All finite iterators are necessarily productive.
...
+make any assumptions about the underlying iterator. In particular, such functions may run forever for certain iterators.
 
-{docstring Iter.allowNontermination}
+Sometimes, it is of utmost importance that a function does terminate.
+For these cases, {name}`Iter.allowNontermination` provides variants of consumers that are guaranteed to terminate.
```
david-christiansen: This is a very surprising API. Should this be this instead?
```suggestion
For these cases, {name}`Iter.ensureTermination` provides variants of consumers that are guaranteed to terminate.
```

Manual/Iterators.lean
```diff
@@ -300,6 +304,18 @@ instance [Pure m] : Iterator Nats m Nat where
     pure <| .deflate <|
       .yield { it with internalState.next := n + 1 } n (by grind)
 ```
+
+Whenever an iterator is defined, {name}`IteratorCollect` and {name}`IteratorLoop` instances should be provided.
+They are required for most consumers of iterators such as {name}`Iter.toList` or the `for` loops.
+One can use their default implementations as follows:
```
david-christiansen: I think this gives a bit more context:
```suggestion
While it is possible to provide specially-optimized implementations for some iterators, most iterators should use the default implementations:
```

Manual/Iterators.lean
```diff
@@ -329,7 +345,18 @@ def Nats.iter : Iter (α := Nats) Nat :=
 :::
 
 :::paragraph
-This iterator is useful with combinators such as {name}`Iter.zip`:
+One can print all natural numbers by running the following function:
```
david-christiansen: While `IO` is in some sense always a function due to its encoding as a state monad behind the scenes, we usually write at a level of abstraction that treats it as opaque. There's no arrow in `f`'s  type, so this should be something like:
```suggestion
This {name}`IO` action prints all the natural numbers:
```

Manual/Iterators.lean
```diff
@@ -329,7 +345,18 @@ def Nats.iter : Iter (α := Nats) Nat :=
...
+```lean
+def f : IO Unit := do
+  for x in Nats.iter do
+    IO.println s!"{x}"
+```
+This function never terminates, printing all natural numbers in increasing order, one
+after another.
```
david-christiansen:
```suggestion
A call to `f` never terminates; it continues printing all natural numbers in increasing order, one
after another.
```

Manual/Iterators.lean
```diff
@@ -342,6 +369,44 @@ This iterator is useful with combinators such as {name}`Iter.zip`:
 2: pachycephalosaurus
 ```
 :::
+
+:::paragraph
+In contrast to the previous example, this loop terminates because `xs.iter` is a finite iterator,
+One can make sure that a loop actually terminates by providing a {name}`Finite` instance:
+```lean (name := natfin)
+#check type_of% (Nats.iter.zip ["cat", "dog"].iter).internalState
```
david-christiansen: This is a nice trick! We don't describe `type_of%` anywhere yet. Should we think of it as a public user-facing feature? Maybe we should add an option to `#check` that suppresses the term, kind of like how `#reduce` has options to reduce in types/proofs?

Manual/Iterators.lean
```diff
@@ -342,6 +369,44 @@ This iterator is useful with combinators such as {name}`Iter.zip`:
 2: pachycephalosaurus
 ```
 :::
+
+:::paragraph
+In contrast to the previous example, this loop terminates because `xs.iter` is a finite iterator,
```
david-christiansen: This sentence should be part of the preceding paragraph, right?

Also:
```suggestion
In contrast to {name}`f`, this loop terminates because `xs.iter` is a finite iterator.
```

Manual/Iterators.lean
```diff
@@ -342,6 +369,44 @@ This iterator is useful with combinators such as {name}`Iter.zip`:
...
+One can make sure that a loop actually terminates by providing a {name}`Finite` instance:
```
david-christiansen:
```suggestion
A {name}`Finite` instance ensures that a loop actually terminates:
```

Manual/Iterators.lean
david-christiansen: This should work, with suitable text adjustment, right?
```suggestion
example : (Zip Nats Id (ListIterator String) String) Id :=
  (Nats.iter.zip ["cat", "dog"].iter).internalState
```

Manual/Iterators.lean
```diff
@@ -342,6 +369,44 @@ This iterator is useful with combinators such as {name}`Iter.zip`:
...
+#synth Finite Nats Id
+```
+```leanOutput natinf
+failed to synthesize
+  Finite Nats Id
+
+Hint: Additional diagnostic information may be available using the `set_option diagnostics true` command.
+```
+If it is important that a loop terminates, {name}`Iter.ensureTermination` can be used. It has a `ForIn` instance
+and other consumers functions that fail if no {name}`Finite` instance can be synthesized.
```
david-christiansen, line 399: What does the second sentence mean here?

This is also a repetition of a sentence from a bit earlier. Perhaps this example should have text that's a bit more focused, with this kind of general description in the main reference text? Maybe something like:

> Calling {name}`Iter.ensureTermination` on `Nats` results in an iterator that cannot be consumed by a {keywordOf Lean.Parser.Term.doFor}`for` loop without a `Finite` instance:

Manual/Iterators.lean
```diff
@@ -441,12 +506,30 @@ def abc : Triple Char := ⟨'a', 'b', 'c'⟩
...
+With the {name}`IteratorCollect` instance in place, {name}`Iter.toArray` now works:
+```lean (name := abcToArray)
+#eval abc.iter.toArray
+```
+```leanOutput abcToArray
+#['a', 'b', 'c']
+```
+
+In general, `Iter.toArray` might run forever. One can prove that `abc` is finite, and the above example will terminate after finitely many steps, by
+constructing a `Finite (Triple Char) Id` instance.
```
david-christiansen:
```suggestion
In general, `Iter.toArray` might run forever.
To use `abc` in a context where termination is important, it must be proved finite by constructing a `Finite (Triple Char) Id` instance.
```

Manual/Iterators.lean
```diff
@@ -487,28 +570,8 @@ instance [Pure m] : Finite (TripleIterator α) m where
...
-{name}`Iter.toArray` now works:
-```lean (name := abcToArray)
-#eval abc.iter.toArray
-```
-```leanOutput abcToArray
-#['a', 'b', 'c']
-```
-
-To enable the iterator in {keywordOf Lean.Parser.Term.doFor}`for` loops, instances of {name}`IteratorLoopPartial` and {name}`IteratorLoop` are needed:
+To enable the iterator in {keywordOf Lean.Parser.Term.doFor}`for` loops, an instance of {name}`IteratorLoop` are needed:
```
david-christiansen:
```suggestion
To enable the iterator in {keywordOf Lean.Parser.Term.doFor}`for` loops, an instance of {name}`IteratorLoop` is needed:
```

Manual/Iterators.lean
```diff
@@ -701,25 +756,26 @@ import Std.Data.Iterators
 ```lean -show
 open Std.Iterators
 ```
-Attempting to construct a list of all the natural numbers from an iterator fails:
-```lean (name := toListInf) +error -keep
+Attempting to construct a list of all the natural numbers from an iterator will produce an endless loop:
```
david-christiansen:
```suggestion
Attempting to construct a list of all the natural numbers from an iterator will produce an infinite loop:
```

Manual/Iterators.lean
```diff
@@ -701,25 +756,26 @@ import Std.Data.Iterators
...
+```lean (name := toListInf) -keep
 def allNats : List Nat :=
   let steps : Iter Nat := (0...*).iter
   steps.toList
 ```
+If endless loops are not desired, one can use the operations available via {lean}`Iter.ensureTermination`. These
+are guaranteed to terminate after finitely many steps, but they will fail if Lean cannot prove the iterator finite.
```
david-christiansen: Is this a correct modification? Or does it need to be more nuanced and mention `Productive` too?
```suggestion
The combinator {lean}`Iter.ensureTermination` results in an iterator for which potentially-infinite operations require evidence of terminations.
In particular, these operations are guaranteed to terminate after finitely many steps, but they require an instance of {name}`Finite` for the iterator.
```

Manual/Iterators.lean
david-christiansen: I figured it out. Never mind :-)

PR #704, opened 2025-12-15
refactor: move error explanations into the manual

robsimmons:
This PR makes error explanations in the reference manual normal Verso pages that use Verso syntax. 

The `Block.tabbedMWEs` block extension in Manual.Meta.ErrorExplanation was recreated as ` ` in Manual.Meta.ErrorExplanation.Example where it is associated with the new `errorExample` directive, and the `Block.errorExplanationMetadata` in Manual.Meta.ErrorExplanation was recreated as `Block.errorExplanationHeader` in Manual.Meta.ErrorExplanation.Header where it is associated with the new `errorExplanationHeader` block command.

### Visual changes:
* Error explanations have links to names, syntax, as appropriate (e.g. `` `match` `` becomes ``{keywordOf Lean.Parser.Term.match}`match` ``<br/><br/>
* Links in the summary table are regular links, and formatted as standard links (old is left, new is right) image<br/><br/>
* Titles are modified from `` `lean.whatever` `` to `` `About: `whatever` `` (which was the consensus as I understand it from [this thread in the FRO zulip](https://lean-fro.zulipchat.com/#narrow/channel/532109-verso/topic/Unknown.20identifier/near/547201770)[#verso > Unknown identifier @ 💬](https://lean-fro.zulipchat.com/#narrow/channel/532109-verso/topic/Unknown.20identifier/near/547201770)), and there's a new line in the header that contains the full error code (old is left, new is right) image<br/><br/>
* Specialized error explanation links elsewhere are now displayed as normal links, which changes their formatting from gray-underline link to regular HTML color link. (old is left, new is right)  image

### Outstanding issues:
* ~Reusing the existing elaborator for lean code fences uncovered what I think is a bug with the way verso code fences handle unparsed content, which causes a problem for [this error explanation](https://lean-lang.org/doc/reference/latest/Error-Explanations/lean___inductionWithNoAlts/#The-Lean-Language-Reference--Error-Explanations--lean___inductionWithNoAlts). Needs to be fixed with leanprover/verso#677.~<br/><br/>
* ~The hover-over display of proof states in `lean` code fences conflicts with the tabbed-example CSS used for error information, and the result is visually confusing. @david-christiansen is there a way to configure that away for existing `lean` code fences, or should this be added? (old without hover-over proofstate pills is left, new with undesirable hover-over proofstate pills is right)~

### DIFF
--- a/Manual/ErrorExplanations/InductionWithNoAlts.lean
+++ b/Manual/ErrorExplanations/InductionWithNoAlts.lean
@@ -33,7 +33,7 @@ theorem zero_mul (m : Nat) : 0 * m = 0 := by
   rw [n_ih]
 ```
 ```output
-unknown tactic
+Invalid syntax for induction tactic: The `with` keyword must followed by a tactic or by an alternative (e.g. `| zero =>`), but here it is followed by the identifier `n`.
 ```
 ```fixed
 theorem zero_mul (m : Nat) : 0 * m = 0 := by


Manual/ErrorExplanations/README.md
```diff
@@ -0,0 +1,124 @@
...
+The reference manual should only need to contain docs for `lean.*` error
+names.
+
+## Error Explanations in the Compiler
+
+New error explanations are declared with the `register_error_explanation` 
+command and take a `Lean.ErrorExplanation.Metadata` structure
+describing the error. All errors have two-component names: the first
+identifies the package or "domain" to which the error belongs (in
+core, this will nearly always be `lean`); the second identifies the
```
david-christiansen:
```suggestion
core, this will always be `lean`); the second identifies the
```
Right?

Manual/ErrorExplanations/README.md
```diff
@@ -0,0 +1,124 @@
...
+%%%
+shortTitle := "foo"
+%%%
+
+{errorExplanationHeader lean.foo}
+
+...mandatory short description...
+
+# Examples
+:::errorExample "Headline Cased Description"
```
david-christiansen: My inner pedant really wants some hyphenation here!

Manual/ErrorExplanations/README.md
```diff
@@ -0,0 +1,124 @@
+# Writing Error Explanations
+
+Error explanations give more context and discussion than is reasonable to
+include in an actual error message. The error message in the Lean InfoView
+contains a link to the error explanation for further reading.
+
+An error explanation is just a manual page obeying a few conventions:
```
david-christiansen, line 7: Do we validate this anywhere?

If not, I can help write a quick check for these conventions that runs in CI after the holidays.

Manual/ErrorExplanations/RedundantMatchAlt.lean
```diff
@@ -0,0 +1,121 @@
...
+
+This error may appear in any pattern matching expression, including
+{keywordOf Lean.Parser.Term.match}`match` expressions, equational function definitions, `if let`
+bindings, and monadic {keywordOf Lean.Parser.Term.let}`let` bindings with fallback clauses.
+
+In pattern-matches with multiple arms, this error may occur if a less-specific pattern precedes a
+more-specific one that it subsumes. Bear in mind that expressions are matched against patterns from
+top to bottom, so specific patterns should precede generic ones.
+
+In `if let` bindings and monadic {keywordOf Lean.Parser.Term.let}`let` bindings with fallback
```
david-christiansen:
```suggestion
In {keywordOf termIfLet}`if let` bindings and monadic {keywordOf Lean.Parser.Term.let}`let` bindings with fallback
```

Manual/ErrorExplanations/RedundantMatchAlt.lean
```diff
@@ -0,0 +1,121 @@
...
+pattern-matching {keywordOf Lean.Parser.Term.let}`let`.
+
+One common cause of this error is that a pattern that was intended to match a constructor was
+instead interpreted as a variable binding. This occurs, for instance, if a constructor
+name (e.g., `cons`) is written without its prefix ({name}`List`) outside of that type's namespace.
+The constructor-name-as-variable linter, enabled by default, will display a warning on any variable
+patterns that resemble constructor names.
+
+This error nearly always indicates an issue with the code where it appears. If needed, however,
+`set_option match.ignoreUnusedAlts true` will disable the check for this error and allow pattern
```
david-christiansen, line 44: We should add a role for `set_option` examples that hyperlinks things correctly. Verso docstrings have one already. I think it's fine to open this as an issue, though, and merge without having links or validation for these for now.

Manual/ErrorExplanations/SynthInstanceFailed.lean
```diff
@@ -0,0 +1,131 @@
...
+
+{errorExplanationHeader lean.synthInstanceFailed}
+
+```lean -show
+variable {t : Type}
+```
+
+{ref "type-classes"}[Type classes] are the mechanism that Lean and many other
+programming languages use to handle overloaded operations. The code that handles a particular
+overloaded operation is an _instance_ of a typeclass; deciding which instance to use for a given
```
david-christiansen:
```suggestion
overloaded operation is an {tech}_instance_ of a typeclass; deciding which instance to use for a given
```
This links to the definition.

Manual/ErrorExplanations/SynthInstanceFailed.lean
```diff
@@ -0,0 +1,131 @@
...
+overloaded operation is called _synthesizing_ an instance.
+
+As an example, when Lean encounters an expression `x + y` where `x` and `y` both have type
```
david-christiansen: It would be nice to highlight this code as well.

Manual/ErrorExplanations/UnknownIdentifier.lean
```diff
@@ -0,0 +1,239 @@
...
+public section
+
+def inventory :=
+  Std.HashSet.ofList [("apples", 3), ("bananas", 4)]
+```
+The constant `Std.HashSet.ofList` is defined in the `Std.Data.HashSet.Basic` module, which has not
+been imported in the original example. This import is suggested by the unknown identifier code
+action; once it is added, this example compiles.
+:::
+-/
```
david-christiansen: Verso doesn't support Lean comment syntax. What's going on here?

Manual/Meta/ErrorExplanation/Example.lean
```diff
@@ -0,0 +1,244 @@
...
+Some explanatory text here.
+:::
+`````
+-/
+@[directive]
+def errorExample : Verso.Doc.Elab.DirectiveExpanderOf ErrorExampleConfig
+  | { title }, contents => do
+    let brokenStx :: restStx := contents.toList
+      | throwError m!"The error example had no contents"
+    let `(Lean.Doc.Syntax.codeblock|``` broken| $brokenTxt ```) := brokenStx
```
david-christiansen, line 196: While this totally works, it is very, very strict and doesn't participate in the usual namespacing rules.

To allow name resolution to work the right way, it might be worth instead realizing whatever the identifier the user wrote, then checking for equality with a built-in that throws errors when used outside. That'll avoid violating user expectations about names.

This is also a tiny detail and I understand if you think it's below the noise floor.

Manual/Meta/ErrorExplanation/Header.lean
```diff
@@ -0,0 +1,124 @@
...
+    }}
+
+structure ErrorHeaderConfig where
+  name : Name
+instance : Verso.ArgParse.FromArgs ErrorHeaderConfig m where
+  fromArgs :=
+    ErrorHeaderConfig.mk <$> Verso.ArgParse.positional `title Verso.ArgParse.ValDesc.name
+
+/--
+Error explanations must start with header inserted by the {lit}`{errorExplanationHeader}` block
```
david-christiansen:
```suggestion
Error explanations must start with a header inserted by the {lit}`{errorExplanationHeader}` block
```

Manual/ErrorExplanations/README.md
robsimmons: We don't — and I'm not sure if we actually want to enforce such a convention for now, given that we might want to change the convention. The structure of an error example is enforced rigorously, and that's more important imo.

Manual/ErrorExplanations/RedundantMatchAlt.lean
robsimmons: Filed a not-very-verbose issue at #712

Manual/Meta/ErrorExplanation/Example.lean
robsimmons: I think I'm comfortable doing an expectation-defying thing here — this is still less fragile than the status quo, and my intention was to be visibly using Metaprogramming Nonsense™️ to treat the codeblocks as syntactic configuration parts of the ```:::errorExample``` rather than as standalone code fences that are usually expected.

Manual/ErrorExplanations/UnknownIdentifier.lean
robsimmons: Filed as leanprover/verso#678

Manual/ErrorExplanations/SynthInstanceFailed.lean
robsimmons: fixed in [45b9492](https://github.com/leanprover/reference-manual/pull/704/commits/45b9492916dabf0071f0584f7fadfbd664ec463a)

Manual/ErrorExplanations/README.md
robsimmons: fixed in [b2112da](https://github.com/leanprover/reference-manual/pull/704/commits/b2112da6e254f0bd3a8290d6288de0b87f5fece8)

PR #711, opened 2025-12-17
chore: enforce autoformatting with prettier

robsimmons:
This PR replays the addition of prettier formatting from leanprover/verso#659 for the reference manual.


david-christiansen:
Did something go wrong? I thought all the reformatting was merged already, but this PR is showing a big reformat along with the action.

Otherwise, great stuff! We need to make sure to add this to the merge queue requirements once it's fully landed.

PR #722, opened 2025-12-27
feat: add a 'highInferredUniverse' error explanation

robsimmons:
This PR introduces a new error explanation for possibly-too-high inferred universes, with the goal of shortening a long but important error explanation. (The theory is that an example is worth several dozen words.)

### DIFF
--- a/Manual/ErrorExplanations/HighInferredUniverse.lean
+++ b/Manual/ErrorExplanations/HighInferredUniverse.lean
@@ -39,13 +39,14 @@ details on the relationship between {lean}`Prop`, {lean}`Sort`, and {lean}`Type`
 :::
 ::::
 
-There are certain situations where universe inference yields
-a universe level that may be higher than it could be.
-In these situations, Lean chooses to avoid silently accepting
-this unnecessarily high universe level, which might cause hard-to-diagnose errors later.
-Avoiding these situations requires an explicitly provided universe for the inductive type
-using a universe level in a slightly different form;
-this cannot be corrected by the elaborator on its own and requires user intervention.
+There are certain situations where universe inference yields a universe level that may be higher
+than it could be.
+In these situations, Lean chooses to avoid silently accepting this unnecessarily high universe
+level, because unnecessarily high universe levels can cause hard-to-diagnose errors in later
+definitions.
+Avoiding these situations requires an explicitly provided universe for the inductive type using a
+universe level in a slightly different form.
+This can't currently be done automatically, so fixing it requires user intervention.
 
 # Examples
 :::errorExample "Use of `Type` May Force Universe Too High"


Manual/ErrorExplanations/HighInferredUniverse.lean
```diff
@@ -0,0 +1,128 @@
...
+
+#doc (Manual) "About: `highInferredUniverse`" =>
+%%%
+shortTitle := "highInferredUniverse"
+%%%
+
+{errorExplanationHeader lean.highInferredUniverse}
+
+This error message appears when Lean's elaboration algorithm is about to assign a new type to a
+higher universe than may be necessary.
```
kmill, line 21: Maybe something like this?
```suggestion
This error message appears when the inferred universe for an `inductive` or `structure`
may be higher than necessary, which may be unintended. Providing an explicit universe
for the type signals user intent and suppresses the error.
```

Manual/ErrorExplanations/HighInferredUniverse.lean
```diff
@@ -0,0 +1,128 @@
...
+:::keepEnv
+```lean
+inductive MaybeTwo (α : Type u)
+  | none
+  | some (fst snd : α)
+```
+```lean -show
+variable (α : Type u)
+```
+Lean must classify {lean}`MaybeTwo α` as a {lean}`Type u`, or, equivalently, as a
```
kmill:
```suggestion
then the universe for {lean}`MaybeTwo α` must be {lean}`Type u` or higher.
Lean automatically infers the universe to be the least universe, hence it chooses {lean}`Type u`, or, equivalently,
```
I'd suggest using "the universe for ... is ..." or "the type of ... is ..." over "classify", unless "classify" has been defined in the manual already for this.

Manual/ErrorExplanations/HighInferredUniverse.lean
```diff
@@ -0,0 +1,128 @@
...
+{lean}`Sort (u + 1)`. (See the reference guide to {ref "universes-sorts"}[Universes] for more
+details on the relationship between {lean}`Prop`, {lean}`Sort`, and {lean}`Type`.)
+:::
+::::
+
+Lean chooses to avoid silently assigning a new type to a universe that might cause confusion later.
+This error often indicates that the universe annotations need to be in a slightly different form,
+or that more specific annotations are required.
```
kmill: Perhaps
```suggestion
There are certain situations where universe inference yields
a universe level that may be higher than it could be.
In these situations, Lean chooses to avoid silently accepting
this unnecessarily high universe level, which might cause hard-to-diagnose errors later.
Avoiding these situations requires an explicitly provided universe for the inductive type
using a universe level in a slightly different form;
this cannot be corrected by the elaborator on its own and requires user intervention.
```

Manual/ErrorExplanations/HighInferredUniverse.lean
```diff
@@ -0,0 +1,128 @@
...
+```
+```fixed
+inductive Bar : Type u
+| foobar {Foo : Sort u} : Foo → Bar
+| barbar : Option Bar → Bar
+```
+
+In this case, the error message is a false positive, but the issue can be easily fixed with
+clarifying annotations. After annotating Foo with a universe `u`, the overall type can be assigned
+type that the error message suggests, which silences the error.
```
kmill, line 133: I'm thinking of modifying the universe level inference algorithm to use a different error here, a simple universe level metavariable error localized to the binder. I think we talked a bit about how it's weird that it creates universe level parameters for constructor fields — this isn't something that should be done automatically.

Manual/ErrorExplanations/HighInferredUniverse.lean
robsimmons: Went with 

```
This error message appears when the inferred universe for an `inductive` or `structure` definition
might be unnecessarily or unintentionally high. In some cases, a straightforward change can lower
the inferred universe. In other cases, providing an explicit universe for the type to make
the user's intent explicit is the right way to remove this error.
```

Manual/ErrorExplanations/HighInferredUniverse.lean
robsimmons: I think that's great — error explanation pages can be valuable even or especially if the error message sometimes appears in wonky places. If you change the error message in this kind of example, then it'll make sense to just remove the example from the manual page.

PR #732, opened 2025-12-30
doc: fix lake tooling wording

alok:
Draft typo/grammar cleanup.


Manual/Meta/LakeToml/Test.lean
```diff
@@ -12,7 +12,7 @@ namespace Manual.Toml
 open Std (Format)
 open Lean Elab
 
-/-- Types that can be used in in-manual tests for TOML decoding -/
+/-- Types that can be used in tests embedded in the manual for TOML decoding -/
```
david-christiansen, line 15: I don't think this is a correction, per se, but it's also fine :)

PR #737, opened 2026-01-06
fix: change 'def' to 'abbrev' in the abbrev section

viol37:
This PR changes `def` to `abbrev` in the section on the `abbrev` command.


david-christiansen:
Thank you!

PR #738, opened 2026-01-06
fix: update recursion techniques count

viol37:
This PR adjusts the paragraph on elaboration of recursive functions to include `partial_fixpoint`, following the changes in https://github.com/leanprover/reference-manual/pull/253.


david-christiansen:
Thanks!

PR #747, opened 2026-01-14
feat: add highlights for v4.27

viol37:
This PR adds highlights section to Lean v4.27 release notes.

### DIFF
--- a/Manual/Releases/v4_27_0.lean
+++ b/Manual/Releases/v4_27_0.lean
@@ -230,11 +230,9 @@ These updates include some *breaking changes*, namely:
   form `x = f (g x)` and stack overflows, which can be fixed by
   generalizing over `g x`.
 
-# Changelog
+# Language
 
 ````markdown
-
-## Language
 
 * [#7965](https://github.com/leanprover/lean4/pull/7965) lets recursive functions defined by well-founded recursion use a
   different `fix` function when the termination measure is of type `Nat`.
@@ -494,7 +492,11 @@ These updates include some *breaking changes*, namely:
 * [#11657](https://github.com/leanprover/lean4/pull/11657) improves upon #11652 by keeping the kernel-reduction-optimized
   definition.
 
-## Library
+````
+
+# Library
+
+```markdown
 
 * [#8406](https://github.com/leanprover/lean4/pull/8406) adds lemmas of the form `getElem_swapIfInBounds*` and deprecates
   `getElem_swap'`.
@@ -748,7 +750,11 @@ These updates include some *breaking changes*, namely:
 * [#11654](https://github.com/leanprover/lean4/pull/11654) updates the `grind` docstring. It was still mentioning `cutsat`
   which has been renamed to `lia`. This issue was reported during ItaLean.
 
-## Tactics
+```
+
+# Tactics
+
+````markdown
 
 * [#11226](https://github.com/leanprover/lean4/pull/11226) finally removes the old `grind` framework `SearchM`. It has been
   replaced with the new `Action` framework.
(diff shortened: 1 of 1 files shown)


Manual/Releases/v4_27_0.lean
```diff
@@ -21,6 +21,208 @@ file := "v4.27.0"
...
 For this release, 372 changes landed. In addition to the 118 feature additions and 71 fixes listed below there were 28 refactoring changes, 13 documentation improvements, 25 performance improvements, 6 improvements to the test suite and 111 other changes.
 
+## Highlights
+
+### Module System Stabilized
+
+[#11637](https://github.com/leanprover/lean4/pull/11637) declares the module system as no longer experimental and makes
+the `experimental.module` option a no-op.
+
+See the [Sources Files and Modules](https://lean-lang.org/doc/reference/latest/Source-Files-and-Modules/#module-scopes)
```
nomeata: This should probably be a verso reference that links to the section in the current document. I don't know if it’s easy to include them in a ```` ```markdown ```` block. @david-christiansen ?

Manual/Releases/v4_27_0.lean
```diff
@@ -21,6 +21,208 @@ file := "v4.27.0"
...
+  equalities (assuming propositional equalities between the indices). This
+  lays ground work for better support for applying injection to
+  heterogeneous equalities in grind.
+  This is a breaking change for whoever uses the `noConfusion` principle
+  manually and explicitly for a type with indices.
+
+- [11562](https://github.com/leanprover/lean4/pull/11562) makes the `noConfusion` principles even more heterogeneous, by
+  allowing not just indices but also parameters to differ.
+  This is a breaking change for manual use of `noConfusion` for types with parameters.
+  Pass suitable `rfl` arguments, and use `eq_of_heq` on the resulting equalities as needed.
```
nomeata:
```suggestion
- [#11474](https://github.com/leanprover/lean4/pull/11474) and 
  [11562](https://github.com/leanprover/lean4/pull/11562)
  generalizes the `noConfusion` constructions to heterogeneous
  equalities (assuming propositional equalities between parameters and indices).   This is a breaking change for whoever uses the `noConfusion` principle
  manually and explicitly for a type with indices
  Pass suitable `rfl` arguments, and use `eq_of_heq` on the resulting equalities as needed.
```

Manual/Releases/v4_27_0.lean
viol37: I’ve moved the highlights out of the markdown block to add the reference. Is that ok? Overall, I think it's rendered nicer this way.

nomeata: Sure, the markdown block is only for convenience.

Manual/Releases/v4_27_0.lean
```diff
@@ -5,22 +5,235 @@ Author: Anne Baanen
...
+- [#11490](https://github.com/leanprover/lean4/pull/11490) prevents `try` swallowing heartbeat errors from nested `simp`
+  calls, and more generally ensures the `isRuntime` flag is propagated by
+  `throwNestedTacticEx`. This prevents the behavior of proofs (especially
+  those using `aesop`) being affected by the current recursion depth or
+  heartbeat limit.
+  This breaks a single caller in Mathlib where `simp` uses a lemma of the
+  form `x = f (g x)` and stack overflows, which can be fixed by
+  generalizing over `g x`.
+
+# Changelog
```
nomeata: The TOC levels differ now from the usual. Is it possible to omit this here and remove one `#` for more consistency with previous release notes?

nomeata: Or if that doesn't work: break this markdown block into many, one per second

Manual/Releases/v4_27_0.lean
david-christiansen: This sounds like a good way to do it :-)

Manual/Releases/v4_27_0.lean
david-christiansen: IIRC the Markdown block attempts to nest headers into whatever context it's used in. I think another good fix would be to have an optional parameter to the block to specify its basic nesting level relative to the Verso document. But for right now, splitting it is likely easiest.

viol37: sure! [5e967b8](https://github.com/leanprover/reference-manual/pull/747/commits/5e967b85e75c3f3f58a3c3bf485fcab24b57c739)

PR #806, opened 2026-03-16
feat: add highlights for v4.29

viol37:
This PR adds highlights section to Lean v4.29 release notes.


Manual/Releases/v4_29_0.lean
```diff
@@ -21,6 +21,308 @@ file := "v4.29.0"
...
+terms step-by-step according to call-by-value semantics. At each
+reduction step, `cbv` produces an equality proof, allowing it to
+certify reductions of functions defined by well-founded recursion or
+partial fixpoints. Unlike {tactic}`native_decide`, the resulting proofs do not
+rely on the compiler. `cbv` is available as a standalone tactic
+([#12408](https://github.com/leanprover/lean4/pull/12408)) and inside
+`conv` mode
+([#12279](https://github.com/leanprover/lean4/pull/12279)). A
+finishing variant, {tactic}`decide_cbv`, is also provided
+([#12411](https://github.com/leanprover/lean4/pull/12411)).
```
nomeata: @wkrozowski , do you think it makes sense to highlight this in 4.29 already, or should we be silent on that and add it to the highlights when it’s no longer experimental?

Manual/Releases/v4_29_0.lean
```diff
@@ -21,6 +21,308 @@ file := "v4.29.0"
...
+
+- `@[cbv_eval]`
+  ([#12296](https://github.com/leanprover/lean4/pull/12296)) and
+  `@[cbv_eval ←]` variant
+  ([#12506](https://github.com/leanprover/lean4/pull/12506)) -
+  register a theorem as a rewrite rule for `cbv`.
+
+## New Extensible `do` Elaborator
+
+[#12459](https://github.com/leanprover/lean4/pull/12459) adds a new,
```
nomeata, line 44: @sgraf812 should we highlight this in 4.29 or only when it becomes the default?

Manual/Releases/v4_29_0.lean
```diff
@@ -21,6 +21,308 @@ file := "v4.29.0"
...
+
+    *Potential regressions*: automation may now behave differently
+    in cases where it previously misidentified instance parameters.
+    For example, a rewrite rule in `simp` that was not firing due to
+    incorrect indexing may now fire.
+
+- [#12340](https://github.com/leanprover/lean4/pull/12340) implements
+  better support for unfolding class fields marked as `reducible`.
+  [#12538](https://github.com/leanprover/lean4/pull/12538) enables
+  this feature.
```
nomeata: This section is important, because it is a serious effort and tricky. Unfortunately, I think it is still in flux (that’s why this time we have this sequence of rcs). I wonder if it makes sense to hold back merging the highlights until we are done, and then write a section that summarizes it all. Do you agree, @kim-em?

But I think we can leave @viol37 off the hook for that.

Manual/Releases/v4_29_0.lean
```diff
@@ -21,6 +21,308 @@ file := "v4.29.0"
 
 For this release, 406 changes landed. In addition to the 107 feature additions and 92 fixes listed below there were 29 refactoring changes, 18 documentation improvements, 28 performance improvements, 26 improvements to the test suite and 104 other changes.
 
+# Highlights
```
nomeata, line 26:
```suggestion
# Highlights

_Violetta Sim has helpfully written the release highlights from 4.16 through 4.29, and the Lean developers gratefully acknowledge her contributions._

```
:-)

Manual/Releases/v4_29_0.lean
wkrozowski: I wouldn't highlight for 4.29 yet, but rather wait until 4.30 (soon planning to turn off the warning about it being experimental, and recently added documentation to the nightly version of the language reference)

Manual/Releases/v4_29_0.lean
sgraf812: I think we can highlight it in 4.29 and then again after it became the default. I want users to try it and play around with it, reporting potential bugs.

Note also that there are breaking changes here to `let pat := rhs | otherwise` syntax that are observable irrespective of whether backward.do.legacy is unset, so if we decide not to highlight we should still acknowledge the breaking change.

nomeata: Ok, feel free to edit this section here as you see fit.

sgraf812: I think it's good as-is 👍

Manual/Releases/v4_29_0.lean
viol37: Ok, I've removed this entry

Manual/Releases/v4_29_0.lean
viol37: thank you! :)

Manual/Releases/v4_29_0.lean
kim-em: Sorry for the slow reply. Yes, I'm looking at this now, and will rewrite this section, and also add some migration advice.

Thanks again @viol37!

Manual/Releases/v4_29_0.lean
```diff
@@ -23,6 +23,344 @@ file := "v4.29.0"
...
+`isDefEq`, and hiding many places where definitional abuse was occurring in downstream libraries.
+
+In order to ensure scalability, and address these definitional abuse problems, we have
+made the quite disruptive change, in [#12179](https://github.com/leanprover/lean4/pull/12179),
+of removing this transparency level bump as the default path.
+
+The changes to the transparency bump in comparing implicit arguments can be controlled in two ways:
+* Definitions can be tagged with the new `@[implicit_reducible]` attribute.
+  This is intermediate between `@[reducible]` and `@[semireducible]` (i.e. the default setting),
+  in that the definition is mostly treated as semireducible, which when `isDefEq` is processing
```
Kha:
```suggestion
  in that the definition is mostly treated as semireducible, except when `isDefEq` is processing
```
?

Manual/Releases/v4_29_0.lean
```diff
@@ -23,6 +23,344 @@ file := "v4.29.0"
...
+  implicit arguments or match discriminants.
+  See [#12247](https://github.com/leanprover/lean4/pull/12247) and [#12567](https://github.com/leanprover/lean4/pull/12567).
+* The option `set_option backward.isDefEq.respectTransparency false` restores the previous behaviour prior to `v4.28.0`
```
Kha:
```suggestion
* The option `set_option backward.isDefEq.respectTransparency false` restores the previous behaviour prior to `v4.29.0`
```

Manual/Releases/v4_29_0.lean
```diff
@@ -23,6 +23,344 @@ file := "v4.29.0"
...
+
+
+As a consequence of these changes to transparency handling, existing definitional abuse problems in downstream libraries now surface in places
+where previously they didn't. To help address these problems, which primarily, but not exclusively,
+are caused by incorrectly implemented typeclass instances, we have made changes in [#12897](https://github.com/leanprover/lean4/pull/12897)
+to `inferInstanceAs` and the default `deriving` handler.
+These ensure that instances created using them do not leak the definition of the types involved,
+when the instances are reduced at less than default transparency.
+
+`inferInstanceAs α` synthesizes an instance of type `α` and then adjusts it to conform to the
```
Kha:
```suggestion
`inferInstanceAs α` synthesizes an instance of type `α` but now adjusts it to conform to the
```
or similar?

Manual/Releases/v4_29_0.lean
```diff
@@ -23,6 +23,344 @@ file := "v4.29.0"
...
+
+Example:
+```
+def D := Nat
+instance : Inhabited D := inferInstanceAs (Inhabited Nat)
+```
+
+The adjustment will make sure that when the resulting instance will not leak the RHS `Nat` when
+reduced at transparency levels below `semireducible`, i.e. where `D` would not be unfolded either,
+preventing “̲defeq abuse”̲.
```
Kha:
```suggestion
reduced at transparency levels below `semireducible`, i.e. where `D` would not be unfolded either.
```
already established using different wording

Kha (dismissed):

kim-em (approved):

PR #839, opened 2026-04-24
feat: improve cache headers for Verso search page

david-christiansen:
This PR adds Netlify header rules that cause the content-addressed parts of the search index to be permanently cached, so clients don't need to do an HTTP round-trip after having downloaded them. A profile in Chrome showed a significant improvement.


david-christiansen:
Don't merge - the search results are returning absolute instead of relative URLs, so this needs fixing first. Investigating...

david-christiansen:
Search result issue fixed upstream in Verso

PR #840, opened 2026-04-29
doc: list reducibility levels from most to least restrictive

kim-em:
Reorder the four reducibility levels in the section on reducibility so the list now reads irreducible → semireducible → implicit reducible → reducible (most restrictive to least), which matches how the levels are usually introduced and aligns with the unfolding hierarchy.

🤖 Prepared with Claude Code


kim-em:
Companion PR to https://github.com/leanprover/lean4/pull/13564. Both came out of reading the docs with @jcommelin.

david-christiansen:
Thanks!

PR #843, opened 2026-05-06
feat: add more tactic examples and documentation

jcreedcmu:


Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+:::example "Providing a Witness"
+To prove that there exists a natural number whose double is 4, it suffices to provide the witness `2`:
+```lean
+example : ∃ n : Nat, n + n = 4 := by
+  exists 2
+```
+:::
+
+The {tactic}`intro` tactic makes progress on goals whose target type is a function type or a universal quantifier.
+It introduces the function's parameter into the local context as a new assumption (for propositions) or a new local variable (for data) and changes the goal to the function's body.
```
david-christiansen: Do we want to call out implication here?

It seems to me that centering function types is the wrong way to present this, and that it'd be better to lead with universal quantifiers, then implications; we're not really in the business of producing computable functions with tactics, but it's fine to have that as the third alternative.

Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+
 :::tactic "intro"
 :::
 
+:::example "Introducing an Implication"
+The goal `P → R` is a function type; {tactic}`intro` moves its premise into the context as `hp`:
```
david-christiansen:
```suggestion
The goal `P → R` is an implication. {tactic}`intro` moves its premise into the context as `hp`:
```

Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+To prove a universally quantified statement, {tactic}`intro` introduces the quantified variable:
+```lean
+example : ∀ (n : Nat), n + 0 = n := by
+  intro n
+  rfl
+```
+:::
+
+:::example "Multiple Introductions"
+Multiple names can be provided to introduce several parameters at once.
```
david-christiansen:
```suggestion
Multiple names can be provided to introduce several assumptions at once.
```

Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+```lean
+example (P Q R : Prop) (h : P → Q → R) :
+    P → Q → R := by
+  intro hP hQ
+  exact h hP hQ
+```
+:::
+
+:::example "Anonymous Introduction"
+When called with no arguments, {tactic}`intro` introduces one parameter using the binder's name if available:
```
david-christiansen: This should mention hygiene (it's really the binder's name plus macro scopes, right?)

Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+```
+:::
+
 
 # Relations
 %%%
 tag := "tactic-ref-relations"
 %%%
 
+The {tactic}`rfl` tactic succeeds whenever the two sides of the relation are {tech (key := "definitional equality")}[definitionally equal], even if they are not syntactically identical.
```
david-christiansen: It also works for other reflexive relations. Should we center that and have an example?

david-christiansen: Oh, there is an example. Mention it here, perhaps?

Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+:::
+
+:::example "Reflexive Relations"
+The {tactic}`rfl` tactic works with any relation that has a lemma tagged with the {attr}`refl` attribute, not just equality.
+For instance, it can close goals involving {lean}`Iff`:
+```lean
+example (P : Prop) : P ↔ P := by
+  rfl
+```
+:::
```
david-christiansen, line 207: What about adding another reflexive relation in an example? Like:

```
def Univ (_ _ : α) := True

@[refl]
theorem Univ.refl (x : α) : Univ x x := True.intro

example : Univ 2 2 := by rfl
```

Manual/Tactics/Reference.lean
```diff
@@ -90,9 +212,21 @@ refl
...
 :::
 
+:::example "Swapping Sides of an Equality"
+After {tactic}`symm`, the goal becomes `a = b`, which matches the hypothesis `h`:
+```lean
+example (a b : Nat) (h : a = b) : b = a := by
+  symm
+  exact h
+```
+:::
```
david-christiansen, line 238: There should also be an example of the symm attribute on a relation here.

Perhaps something like:
```
def Univ (_ _ : α) := True

@[symm]
theorem Univ.symm (x : α) : Univ x y → Univ y x := by
  intro; exact True.intro

example : Univ 2 3 → Univ 3 2 := by
  intro
  symm
  assumption

```

Manual/Tactics/Reference.lean
```diff
@@ -105,28 +239,100 @@ symm
...
+Here `congr 2` peels off two layers of arithmetic operations, leaving `a * c = b * c` as a subgoal that still needs to be solved:
+```lean
+example (a b c : Nat) (h : b * c = a * c) :
+    (a * c) * 2 + 3 = (b * c) * 2 + 3 := by
+  congr 2
+  symm
+  exact h
+```
+Using uncontrolled {tactic}`congr` would have left us with the goal `a = b`,
+which we cannot close, because even though `b * c = a * c`, it might be because `c` is zero.
```
david-christiansen: "close" is not the verb we usually use in this manual. What about "which is unprovable"?

Also, this sentence is long. Can it be two sentences?

Manual/Tactics/Reference.lean
```diff
@@ -150,15 +356,85 @@ tag := "tactic-ref-associativity-commutativity"
 tag := "tactic-ref-lemmas"
 %%%
 
+The {tactic}`exact` tactic closes the current goal by providing a term whose type matches the goal's target type.
```
david-christiansen: Is "closes" the right verb?

Manual/Tactics/Reference.lean
```diff
@@ -150,15 +356,85 @@ tag := "tactic-ref-associativity-commutativity"
...
+
+:::example "Applying Lemmas"
+The argument to {tactic}`apply` is not limited to local hypotheses.
+Any term whose conclusion matches the goal can be used, including lemmas:
+```lean
+example (a b c : Nat) (hab : a < b) (hbc : b < c) :
+    a < c := by
+  apply Nat.lt_trans
+  · exact hab
+  · exact hbc
```
david-christiansen: What about `apply` here, for extra oomph?

Manual/Tactics/Reference.lean
```diff
@@ -150,15 +356,85 @@ tag := "tactic-ref-associativity-commutativity"
...
+```
+:::
+
+Note that {tactic}`apply` does not work directly with `↔` (if-and-only-if) hypotheses.
+To use a hypothesis `h : P ↔ Q` backwards on the goal, use {tactic}`rw` instead, or extract one direction with `h.mp` or `h.mpr`.
```
david-christiansen, line 449: This could use an example.

Manual/Tactics/Reference.lean
```diff
@@ -150,15 +356,85 @@ tag := "tactic-ref-associativity-commutativity"
...
+
+The {tactic}`refine` tactic is like {tactic}`exact`, but allows holes written as `?_` that become new goals.
+This is useful when part of a term is known but some arguments still need to be proved.
```
david-christiansen, line 460: It's also nice when the proof being applied isn't just a lemma that mostly matches, but is built up with things like anonymous constructor syntax.

Manual/Tactics/Reference.lean
```diff
@@ -177,12 +453,42 @@ tag := "tactic-ref-lemmas"
 tag := "tactic-ref-false"
 %%%
 
+The {tactic}`exfalso` tactic changes the goal to `False`. It is named after the Latin phrase “ex falso quodlibet”, that is, “from falsehood, anything follows”.
```
david-christiansen:
```suggestion
The {tactic}`exfalso` tactic changes the goal to `False`. It is named after the Latin phrase _ex falso quodlibet_, that is, “from falsehood, anything follows”.
```

Foreign phrases are usually italicized.

Manual/Tactics/Reference.lean
```diff
@@ -177,12 +453,42 @@ tag := "tactic-ref-lemmas"
...
+This is useful when the hypotheses are contradictory: once the goal is `False`, it can be closed by deriving a contradiction.
```
david-christiansen:
```suggestion
This is useful when the hypotheses are contradictory: once the goal is {name}`False`, it can be closed by deriving a contradiction.
```

Manual/Tactics/Reference.lean
```diff
@@ -177,12 +453,42 @@ tag := "tactic-ref-lemmas"
...
 
+:::example "Closing a Goal by Contradiction"
+{tactic}`contradiction` searches the hypotheses automatically for immediate logical contradictions.
+```lean
+example (hp : P) (hnp : ¬P) : Q := by
+  contradiction
+```
+It also recognizes incompatibilities between constructors of the same type.
+```lean
+example (h : 0 = 1) : P := by
```
david-christiansen: I think this could be punchier without the notation:
```suggestion
example (h : Nat.zero = Nat.succ Nat.zero) : P := by
```
or even
```suggestion
example (h : ["non"] = []) : P := by
```
or
```suggestion
example (h : some "str" = none) : P := by
```

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
 tag := "tactic-ref-goals"
 %%%
 
+With {tactic}`suffices`, the user first shows how a sufficient condition would imply the original goal, and then proves that sufficient condition.
```
david-christiansen:
```suggestion
The {tactic}`suffices` tactic replaces the goal with another statement that is at least as strong.
That is, the new goal suffices to show the old one.
```

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
...
+:::
+
+The {tactic}`change` tactic replaces the goal (or a hypothesis) with a {tech (key := "definitional equality")}[definitionally equal] alternative.
+This can make the goal easier to read or bring it into a form that other tactics expect.
+
 :::tactic "change"
 :::
 
+:::example "Unfolding a Definition"
+Because `¬P` is defined as `P → False`, {tactic}`change` can make this explicit:
```
david-christiansen, line 573: This strikes me as a strange use of `change`, because normally I'd use `rw [Not]` or perhaps `unfold [Not]` for this kind of thing.

What about instead proving that `x + 2 = 1 + x + 1` by first `change`ing it to `x + 1 + 1` and then applying associativity? That seems similarly small, but it matches the pattern of "put it into this other form from the same defeq equivalence class that my solver can deal with"

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
...
+  exact hnp hp
+```
+:::
+
 :::tactic "generalize"
 :::
 
+The {tactic}`specialize` tactic instantiates a universally quantified or function-typed hypothesis with specific arguments, replacing it in the context with the result.
+Because {tactic}`specialize` modifies the hypothesis in place, the original general statement is lost after specialization.
+If the original hypothesis is needed again, use {tactic}`have` to create a copy first, for example `have h' := h` before specializing `h`.
```
david-christiansen: Or use `have` to create the specialized version!

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
...
+```lean
+example (h : ∀ a b : Nat, a + b = b + a) :
+    1 + 2 = 2 + 1 := by
+  specialize h 1
+  exact h 2
+```
+:::
+
+The {tactic}`obtain` tactic is used when a hypothesis or proof term has internal structure that should be broken apart.
+Where {tactic}`have` introduces a single new fact into the context, {tactic}`obtain` destructs a term into its pieces using pattern matching. For example, extracting the witness from an existential or the two sides of a conjunction.
```
david-christiansen, line 612: This also uses the `rcases` pattern language, right?

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
...
+
 :::tactic "obtain"
 :::
 
+:::example "Unpacking an Existential"
+The pattern `⟨n, hn⟩` extracts the witness `n` and the proof `hn : n + n = 10` from the existential hypothesis:
+```lean
+example (h : ∃ n : Nat, n + n = 10) : ∃ m : Nat, m = 5 := by
+  obtain ⟨n, hn⟩ := h
+  exact ⟨n, by omega⟩
```
david-christiansen:
```suggestion
  exact ⟨n, by grind⟩
```

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
...
+:::example "Unpacking a Conjunction"
+The same pattern works for conjunctions:
+```lean
+example (h : P ∧ Q) : Q ∧ P := by
+  obtain ⟨hp, hq⟩ := h
+  exact ⟨hq, hp⟩
+```
+:::
+
+The {tactic}`show` tactic selects a goal whose target unifies with the given type and makes it the current goal.
```
david-christiansen: We actually don't get a lot into unification in these docs. It might be nice to unpack this a bit to increase accessibility for those not well-versed in PL implementation.

Manual/Tactics/Reference.lean
```diff
@@ -192,24 +498,111 @@ tag := "tactic-ref-false"
...
+When there is only one goal, it can be used like {tactic}`change` to restate the goal in a {tech (key := "definitional equality")}[definitionally equal] form.
+
 :::tactic "show"
 :::
 
+:::example "Selecting a Goal"
+When there are multiple goals, {tactic}`show` brings a specific one to the front.
+Here, after {tactic}`constructor` the goals are `⊢ P` then `⊢ Q`, but `show Q` reorders them:
```
david-christiansen, line 648: If unification is important here then we should have an example that uses it more obviously.

Manual/Tactics/Reference.lean
```diff
@@ -260,18 +653,43 @@ They are described in more detail by {citet castPaper}[].
 tag := "tactic-ref-ext"
 %%%
 
+The {tactic}`ext` tactic applies extensionality lemmas registered with the {attr}`ext` attribute.
+The principle of extensionality states that two objects are equal if they are built from the same components. For example, two functions are equal if they return the same value on every input.
```
david-christiansen: I don't think that's what extensionality means. Rather, it means that two terms are equal if they denote the same underlying mathematical object, right? So eta is a kind of extensional equality, and funext is a more general form, but that's because the mathematical community considers "maps equal elements of the domain to equal elements of the range" to be what it means for two functions  to be equal. Similarly, we could create a quotient that collapses Nat into even and odd equivalence classes, and extensionality for that type would be precisely this equivalence class. "Nothing can distinguish them means they're equal" is another way to understand it, which I think makes most sense as an intuition here. That captures funext, but also principles like `Array.ext` that says that the arrays have the same size and assign equal valid indices to equal values, or structure ext theorems that are propositional versions of structure eta.

Manual/Tactics/Reference.lean
```diff
@@ -260,18 +653,43 @@ They are described in more detail by {citet castPaper}[].
...
+```
+:::
+
 :::tactic Lean.Elab.Tactic.Ext.tacticExt1___
 :::
 
 :::tactic Lean.Elab.Tactic.Ext.applyExtTheorem
 :::
 
+The {tactic}`funext` tactic is a variant of {tactic}`ext` that specifically applies function extensionality.
```
david-christiansen, line 740: This should cross-reference the funext section of the quotients chapter

Manual/Tactics/Reference.lean
```diff
@@ -293,9 +711,43 @@ tag := "tactic-ref-ext"
 tag := "tactic-ref-rw"
 %%%
 
+The {tactic}`rw` (rewrite) tactic does rewriting: it substitutes equals for equals.
```
david-christiansen:
```suggestion
The {tactic}`rw` uses proofs of equality to rewrite goals and/or hypotheses, replacing occurrences of one of the equated terms with the other.
```

Manual/Tactics/Reference.lean
david-christiansen: Logical equivalence is also extensional equality for `Prop`, as witnessed by `propext`. That definitely doesn't fit "built from the same components" - `(2 = 2) = True` is a theorem of Lean, for example.

Manual/Tactics/Reference.lean
```diff
@@ -293,9 +711,43 @@ tag := "tactic-ref-ext"
...
+Given a proof of an equality `h : x = y` or an if-and-only-if `h : P ↔ Q`, it replaces occurrences of the left-hand side with the right-hand side in the goal.
+Use `rw [← h]` to rewrite in the reverse direction, and `rw [h] at hyp` to rewrite in a hypothesis `hyp` rather than the goal.
+After rewriting, {tactic}`rw` automatically tries to close the goal with {tactic}`rfl`.
```
david-christiansen: It strikes me that this text really belongs in the docstring rather than just here.

Manual/Tactics/Reference.lean
```diff
@@ -52,28 +64,138 @@ tag := "tactic-ref-assumptions"
...
+example : ∃ n : Nat, n + n = 4 := by
+  exists 2
+```
+:::
+
+The {tactic}`intro` tactic makes progress on goals whose target type is a function type or a universal quantifier.
+It introduces the function's parameter into the local context as a new assumption (for propositions) or a new local variable (for data) and changes the goal to the function's body.
+
 :::tactic "intro"
 :::
```
david-christiansen, line 84: The `tactic` directive shows the docstring in a box, along with whatever other text is in the directive's body.

How have you been thinking about which text belongs around the boxes, and which belongs in the boxes?

Manual/Tactics/Reference.lean
```diff
@@ -293,9 +711,43 @@ tag := "tactic-ref-ext"
...
+
+:::example "Rewriting in a Hypothesis"
+The `at` clause directs the rewrite to a hypothesis instead of the goal:
+```lean
+example (x y : Nat) (h : x = y) (h2 : x + 1 = 3) :
+    y + 1 = 3 := by
+  rw [h] at h2
+  exact h2
+```
+:::
```
david-christiansen, line 810: One of these examples should have multiple rewrites, ideally in both directions. Multiple-rule rewrites are very common in practice.

Manual/Tactics/Reference.lean
```diff
@@ -293,9 +711,43 @@ tag := "tactic-ref-ext"
...
+
+Note that {tactic}`rw` does not rewrite under binders such as `∀`, `∃`, or `∑`.
+For example, if `h : a = b`, then `rw [h]` will not rewrite occurrences of `a` inside `∀ x, f a x = g x`.
```
david-christiansen, line 822: Maybe also xref `conv` here, in case `rw` picks the wrong instance of the LHS?

Manual/Tactics/Reference.lean
```diff
@@ -293,9 +711,43 @@ tag := "tactic-ref-ext"
...
+The `at` clause directs the rewrite to a hypothesis instead of the goal:
+```lean
+example (x y : Nat) (h : x = y) (h2 : x + 1 = 3) :
+    y + 1 = 3 := by
+  rw [h] at h2
+  exact h2
+```
+:::
+
+Note that {tactic}`rw` does not rewrite under binders such as `∀`, `∃`, or `∑`.
```
david-christiansen, line 821: A negative example here would be valuable

Manual/Tactics/Reference.lean
```diff
@@ -336,22 +788,57 @@ tag := "tactic-ref-inductive"
 tag := "tactic-ref-inductive-intro"
 %%%
 
+The {tactic}`constructor` tactic tries to solve the goal by application of a constructor.
+When the goal's target type has a single constructor, it replaces the goal with one subgoal for each of the constructor's arguments.
+This is commonly used to split a goal of the form `P ∧ Q` into separate goals for `P` and `Q`, or to split `P ↔ Q` into the two implications.
+It is essentially syntactic sugar for `refine ⟨?_, ?_, …⟩` with as many holes as the constructor has arguments.
```
david-christiansen: This isn't correct. It picks the first constructor whose type unifies with the goal. Here's some behavior to explain:
```
example : False ∨ True := by
  -- it picked Or.inl because the type unified
  constructor
  sorry

inductive SumTwo : Nat → Nat → Prop where
  | zero : SumTwo 0 2
  | one : SumTwo 1 1
  | two : SumTwo 2 0

example : SumTwo 1 1 := by
  -- it picked SumTwo.one because zero's type didn't unify with the goal
  constructor
```

Manual/Tactics/Reference.lean
```diff
@@ -336,22 +788,57 @@ tag := "tactic-ref-inductive"
...
+```
+:::
 
 :::tactic "injection"
 :::
 
 :::tactic "injections"
 :::
 
+The {tactic}`left` and {tactic}`right` tactics select which side of a disjunction to prove.
```
david-christiansen: These actually work for any two-constructor inductive type, so they work for more than just `Or`

Manual/Tactics/Reference.lean
jcreedcmu: Ah, agree that centering the propositional angle is better.

Manual/Tactics/Reference.lean
```diff
@@ -455,18 +942,88 @@ cases_eliminator
 ```
 :::
 
+The {tactic}`cases` tactic performs case analysis on a term in the local context.
+It decomposes the term according to the constructors of its inductive type, creating one subgoal for each constructor.
```
david-christiansen: Technically it uses the registered cases eliminator, which for most types is just what you're specifying here. But it doesn't have to be - Nat has a special one that rephrases goals in terms of `0` and `n + 1` instead of `Nat.zero` and `Nat.succ n`.

You can see this in action with this example:
```
def Bool' := { n : Nat // n = 0 ∨ n = 1 }

@[match_pattern] def Bool'.true : Bool' := ⟨1, by grind⟩
@[match_pattern] def Bool'.false : Bool' := ⟨0, by grind⟩ 
def Bool'.and : Bool' → Bool' → Bool'
  | .true, .true => .true
  | _, _ => .false

@[cases_eliminator]
def Bool'.cases {motive : Bool' → Sort u} (f : motive .false) (t : motive .true) (b : Bool') : motive b := by
  let ⟨n, ok⟩ := b
  match n with
  | 0 => exact f
  | 1 => exact t

example : Bool'.and x y = .true → x = .true ∧ y = .true := by
  cases x <;> cases y <;> simp [Bool'.and]
```

`Bool'` clearly has only one constructor, but `cases x` and `cases y` follow `Bool'.cases` in creating two cases. What's really going on is that it's creating one proof goal for each minor premise of the registered `cases_eliminator`, which by default is the one generated by the `inductive` machinery from the recursor but doesn't need to be.

Manual/Tactics/Reference.lean
```diff
@@ -455,18 +942,88 @@ cases_eliminator
...
+The {tactic}`cases` tactic can also split data, such as a natural number into the `zero` and `succ` cases:
+```lean
+example (n : Nat) : n = 0 ∨ n ≥ 1 := by
+  cases n with
+  | zero => left; rfl
+  | succ m => right; grind
+```
+:::
+
+The {tactic}`rcases` tactic is a recursive version of {tactic}`cases` that destructs a hypothesis using pattern matching notation.
```
david-christiansen: It's important here that the `rcases` pattern language is not the same one used by `match`, `let`, etc.

Manual/Tactics/Reference.lean
```diff
@@ -455,18 +942,88 @@ cases_eliminator
...
+  rcases h with ⟨hp, hq, hr⟩
+  exact ⟨hr, hq, hp⟩
+```
+:::
+
 :::tactic "fun_cases"
 :::
 
+The {tactic}`induction` tactic performs mathematical induction.
+Like {tactic}`cases`, it splits into cases, but it additionally provides an inductive hypothesis in each recursive case.
```
david-christiansen, line 1105: This can also be rigged up to use a custom eliminator.

Manual/Tactics/Reference.lean
```diff
@@ -455,18 +942,88 @@ cases_eliminator
...
+This makes {tactic}`induction` a useful tool for proving properties of recursive data, especially natural numbers.
+
 :::tactic "induction"
 :::
 
+:::example "Induction on Natural Numbers"
+Here we use induction to establish that zero is the identity for addition on the left.
+The base case `zero` is closed by {tactic}`rfl`, and the successor case uses the inductive hypothesis `ih : 0 + n = n`:
```
david-christiansen: This is technically not following the rules you state above, because it's phrasing the goal in simp normal form instead of in terms of constructors. This is accomplished by having a custom induction_eliminator set up for Nat.

Perhaps `List` is a better example?

david-christiansen: (it's good to at least mention the custom eliminator here for attentive readers)

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
 %%%
 
 
+The {tactic}`split` tactic splits the goal on a match expression or if-then-else, creating one subgoal per branch.
```
david-christiansen, line 1199: These subgoals include hypotheses that rule out the control-flow branches not taken, which means they preserve the "first match wins" semantics of `match`. This is crucial.

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
...
+
 :::tactic "split"
 :::
 
+:::example "Splitting a Match Expression"
+The {tactic}`split` tactic creates one subgoal per branch, allowing each to be proved using the relevant hypotheses:
```
david-christiansen, line 1205: There should be a first-match-wins example here. Something like:
```
def classify (n : Nat) : Option String :=
  match n with
  | 7 => some "seven"
  | 12 => some "twelve"
  | _ => none
```
Using `split`on that will result in goals that include hypotheses that 7 and 12 were not matched. This is a big part of the value of `split`.

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
...
+```lean
+example (p : Bool) (x y : Nat) (hx : x > 0) (hy : y > 0) :
+    (match p with | true => x | false => y) > 0 := by
+  split
+  · exact hx
+  · exact hy
+```
+:::
+
+The {tactic}`by_cases` tactic splits the proof into two cases based on whether a proposition holds or not.
```
david-christiansen: Usually we say "is true" for a proposition.  "holds" isn't wrong, but sometimes it's useful to reserve that for metalanguage statements (ie judgments). In any case, consistency of terminology is useful.

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
...
+Always name the hypothesis with `h : P` syntax; without a name, Lean generates an inaccessible name (`h✝`) that cannot be referred to in the proof.
```
david-christiansen: Why is that so bad? `grind` and other automation does just fine with inaccessible hypothesis names.

An xref to tactic hygiene here would be useful.

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
...
+Here we see how {tactic}`decide` can close simple goals.
+```lean
+example : 2 + 2 = 4 := by decide
+```
+```lean
+example : ¬(3 = 5) := by decide
+```
+:::
+
+Because {tactic}`decide` runs the decision procedure at kernel level, it can be extremely slow or time out on large instances.
```
david-christiansen:
```suggestion
Because {tactic}`decide` runs the decision procedure using the kernel's term reduction, it can be extremely slow or time out on large problems.
```

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
...
+For example, checking `Nat.Prime 104729` with {tactic}`decide` would take impractically long.
```
david-christiansen, line 1278:
```suggestion
For example, checking {lean}`Nat.Prime 104729` with {tactic}`decide` would take impractically long.
```

Manual/Tactics/Reference.lean
```diff
@@ -531,21 +1088,64 @@ tag := "tactic-ref-cases"
...
+For arithmetic goals involving large numbers, {tactic}`grind`, {tactic}`omega` or `norm_num` are more performant.
```
david-christiansen:
```suggestion
For arithmetic goals involving large numbers, {tactic}`grind`, {tactic}`omega` or {tactic}`norm_num` are more performant.
```

Are we still encouraging `omega`? I thought it was firmly in the realm of "for bootstrapping" by this point.

Manual/Tactics/Reference.lean
```diff
@@ -1164,9 +1764,21 @@ tag := "tactic-ref-debug"
 tag := "tactic-ref-other"
 %%%
 
+The {tactic}`trivial` tactic tries a short list of simple tactics, including {tactic}`rfl`, {tactic}`assumption`, and {lean}`True.intro`, to close the goal.
```
david-christiansen: This tactic can be extended, right?

Manual/Tactics/Reference.lean
```diff
@@ -1164,9 +1764,21 @@ tag := "tactic-ref-debug"
...
+
 :::tactic "trivial"
 :::
 
+:::example "Closing Easy Goals"
+The {tactic}`trivial` tactic can close goals that are trivial from the point of view of propositional logic.
```
david-christiansen: Strictly speaking, propositional logic doesn't have a point of view :-)

Manual/Tactics.lean
```diff
@@ -842,12 +842,46 @@ Along with these operators, {tactic}`rename_i` allows inaccessible assumptions t
 :::tactic "rename"
 :::
 
+The {tactic}`revert` tactic is the inverse of {tactic}`intro`: it moves a hypothesis from the local context back into the goal as a premise.
+
 :::tactic "revert"
 :::
 
+:::example "Reverting Before Induction"
+Reverting a hypothesis before {tactic}`induction` gives a stronger inductive hypothesis that quantifies over the reverted variable.
```
david-christiansen, line 851: Normally I'd use `generalizing` here rather than a low-level `revert`-then-`induction` dance.

Manual/Tactics.lean
```diff
@@ -842,12 +842,46 @@ Along with these operators, {tactic}`rename_i` allows inaccessible assumptions t
...
+```lean
+example (n m : Nat) : n + m = m + n := by
+  revert m
+  induction n with
+  | zero => intro m; simp
+  | succ n ih => intro m; rw [Nat.succ_add, ih, Nat.add_succ]
+```
+:::
+
+The {tactic}`clear` tactic removes a hypothesis from the local context. This is sometimes useful to simplify the local context.
```
david-christiansen:
```suggestion
The {tactic}`clear` tactic removes a hypothesis from the local context. 
This is sometimes useful to make proof goals easier to understand or to prevent automated tools from following red herrings.
```

Manual/Tactics.lean
```diff
@@ -842,12 +842,46 @@ Along with these operators, {tactic}`rename_i` allows inaccessible assumptions t
...
+
 :::tactic "clear"
 :::
 
+:::example "Clearing Before Induction"
+A hypothesis that mentions the induction variable can be included in the inductive hypothesis in an undesirable way.
```
david-christiansen:
```suggestion
A hypothesis that mentions the induction variable can be included in the inductive hypothesis, resulting in an induction hypothesis that is too weak.
```

Manual/Tactics/Reference.lean
jcreedcmu: Sgtm.

Manual/Tactics/Reference.lean
jcreedcmu: Going with "prove" here.

Manual/Tactics/Reference.lean
jcreedcmu: Done.

Manual/Tactics/Reference.lean
jcreedcmu: Done.

Manual/Tactics/Reference.lean
jcreedcmu: Good idea.

Manual/Tactics/Reference.lean
jcreedcmu: Good idea!

Manual/Tactics/Reference.lean
jcreedcmu: I don't actually have any clear idea how `intro` interacts with macro scopes. Is there another section of the manual that I should be linking to?

Manual/Tactics/Reference.lean
jcreedcmu: Done.

Manual/Tactics/Reference.lean
jcreedcmu: Feel free to keep pushing back if you want to, but I would have found this example helpful when I was learning to know that `change` can let the user deliberately set the goal to exactly the defeq equivalence class representative they want to, even if there are other ways to use other tactics to get to the same endpoint. But the 2 defeq 1+1 is a fine example as well, and I'd be enthusiastic about including both.

Manual/Tactics/Reference.lean
jcreedcmu: I'm not knowledgeable enough to know what is the best thing to include here to that effect textually.

Manual/Tactics/Reference.lean
jcreedcmu: True! Fixed.

Manual/Tactics/Reference.lean
jcreedcmu: I currently feel like that's another topic out of scope of "improve tactic documentation".

Manual/Tactics/Reference.lean
jcreedcmu: I also feel here this is a thing that would be an improvement in general but I'd prefer to keep it out of the scope of this PR.

Manual/Tactics/Reference.lean
jcreedcmu: Hm here I agree that my wording is insufficent to cover propext, but I do like offering an intuition that works for some cases. Will try to reword...

jcreedcmu: I think appealing to Leibniz equality is too tautologous to really feel to me like it's about *extensionality*. A beta-reduced term has all the properties of the beta-redex, but the reason they are equal is via beta-reduction, not via eta.

Manual/Tactics/Reference.lean
david-christiansen: https://lean-lang.org/doc/reference/latest/find/?domain=Verso.Genre.Manual.section&name=tactic-language-hygiene

david-christiansen: Basically, it does the usual macro introduction thing. If the identifier occurs in the syntax that the user wrote, then it's available for reference in further things the user wrote. If the identifier was produced by metaprogram internals (including "give me a default name from this internal binder name"), then it can't be referenced from things the user wrote. And different metaprograms (and invocations of the same metaprogram) get distinct local namespaces as well, preventing painful capture issues.

Manual/Tactics/Reference.lean
jcreedcmu: How about: "Extensionality properties say that two objects are equal if they are equal under all appropriate observations.
For example, two functions are equal if they return the same value on every input and two pairs are equal if their components are equal."

For me, "appropriate" is doing a lot of work in that sentence, and that's ok :-)
I think this phrasing *apples* to Array.ext and propext, but I'm also ok with those example being silent.

Manual/Tactics/Reference.lean
david-christiansen: Both makes sense. I guess the key is that many defeq rules read like directional reduction steps (beta or delta, for instance), but defeq is an _equality_ and `change` works "backwards" too.

Manual/Tactics/Reference.lean
david-christiansen: Right now, the `rcases` pattern language is described in the [`rcases` docstring](https://lean-lang.org/doc/reference/latest/find/?domain=Verso.Genre.Manual.doc.tactic&name=Lean.Parser.Tactic.rcases). This is not good - it should be described in a manual section and linked to from docstrings. I just haven't gotten to it yet. It would be really useful to extract that to a section that all these can xref.

Manual/Tactics/Reference.lean
david-christiansen: I mean "unpack" as in "use a few more words so this can make sense to nonexperts", not "fully define unification". Like "the proposed goal is matched against available goals modulo defeq and assigning of metas/holes", and then include an example that actually demonstrates this.

Manual/Tactics/Reference.lean
david-christiansen: I think this example is sufficient to demonstrate the key principles:
```
example : 1 ≠ 2 ∧ "two" ≠ "three" := by
  constructor
  show "two" = _ → False
```
Is that really too much for the PR? It demonstrates that the matching is modulo defeq (because inequality is unfolded to negation, and that to implication) and holes (because `_` replaces `"three"`)

Manual/Tactics/Reference.lean
jcreedcmu: I can do this but I don't know the canonical way to get names for these hypotheses. If `split with` is a thing, it seems to only be a mathlib thing. I can use `next` but I'm not stoked about using it so.
```
When splitting a case match, hypotheses are available that show that previous arms of the case did not match.
````lean
def classify (n : Nat) : Option String :=
  match n with
  | 7 => some "seven"
  | 12 => some "twelve"
  | _ => none

example (n : Nat) :
    classify n = some "seven" ∨ classify n = some "twelve" ∨
      (n ≠ 7 ∧ n ≠ 12) := by
  unfold classify
  split
  · left; rfl
  · right; left; rfl
  · right; right; next hx hy => refine ⟨hx, hy⟩
````
```

Manual/Tactics/Reference.lean
jcreedcmu: Sure, happy to change.

Manual/Tactics/Reference.lean
jcreedcmu: Ok, I'm happy to soften the wording.

jcreedcmu: I don't really understand why tactic hygiene is relevant here? This isn't challenging the relevance so much as confessing I'm missing information.

Manual/Tactics/Reference.lean
jcreedcmu: Nor does a mountain *have* one, but one may speak of the point of view of the peak, no? :) In any case that's my intent here. Feel free to suggest another wording if you'd like, I'm not too attached to it.

Manual/Tactics.lean
jcreedcmu: I don't ever use revert, so I didn't have a strong notion of how to make a super unimpeachably high-quality example.

Manual/Tactics/Reference.lean
david-christiansen: Well, they're _intensionally_ equal due to beta, but they can be extensionally equal as well due to something like funext or propext or the eta-rule for unit-like types. We can prove that `id () = ()` by appeal to `rfl` and thus defeq via delta and beta, but we can also prove it via something like `Unit.ext (x : Unit) x = ()` (which we have definitionally in Lean too, but pretend we don't for a moment!) or its `Subsingleton` instance.

I think Leibnitz equality is not really the same here, right? I use Leibnitz equality to mean roughly "indistinguishable things are equal", that is,

```
def LEq (x y : a) : Prop := forall P : a -> Prop, P x = Py
```
whereas extensional equality is something defined per-type that allows us to prove `Eq`, of the form `forall x y : a, EQS x y -> x = y` where `EQS` is some telescope of equalities of `x` and `y` in various contexts. This is usually related to/inspired by the intended semantics of the type in the standard Lean model, so funext captures set-theoretic extensional equality of functions, propext captures the fact that there's classically only really two propositions, ext for a structure captures that it's basically just a product, etc.

david-christiansen: Our comments crossed :-) I think the "all appropriate observations" is a fine way to describe it for this context.

Manual/Tactics/Reference.lean
david-christiansen: You can name them yourself with `rename_i`. But I think it's sufficient to state the proposition in the hypothesis without giving it a name here, like:

> In the second case, `split` inserts the hypothesis `n \ne 7` to indicate that the first case in `classify` was not taken. In the third case, it inserts both `n \ne 7` and `n \ne 12`, indicating that neither the first nor the second case matched.

Manual/Tactics/Reference.lean
david-christiansen: If you don't name the hypothesis, then a name is assigned automatically. Due to hygiene, that name is inaccessible, so you can't refer to it later in the proof (but automation can still use it).

Manual/Tactics/Reference.lean
david-christiansen: "goals that are sufficiently easy to prove" perhaps?

Manual/Tactics.lean
david-christiansen: Same here. It is a kind of low-level thing that I expect I'd only really use in a tactic macro or something. I'd just leave the example alone.

PR #849, opened 2026-05-11
doc: clarify what checks honest library authors must do

### LINKED ISSUE #848: Unclear phrasing in ValidatingProofs.lean
It is unclear what check exactly the phrase "[that they performed this check](https://github.com/leanprover/reference-manual/blob/main/Manual/ValidatingProofs.lean#L67C188-L67C207)" refers to. My understanding is that it should be rephrased to "that they checked that the theorems in their libraries express their intended informal meanings" but I'm not sure.

OrfeasLitos:
The previous phrasing "[that they performed this check](https://github.com/leanprover/reference-manual/blob/main/Manual/ValidatingProofs.lean#L67C188-L67C207)" was ambiguous. The new phrasing explicitly states who should do what check.

Closes #848

### DIFF
--- a/Manual/ValidatingProofs.lean
+++ b/Manual/ValidatingProofs.lean
@@ -64,7 +64,7 @@ The blue ticks indicate that the theorem statement has been successfully elabora
 
 ## Trust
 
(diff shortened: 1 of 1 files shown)


nomeata (approved):

OrfeasLitos:
Fixed commit message as per convention and PR title & description to match.

PR #855, opened 2026-05-27
doc: expand the lake test and lint driver section

kim-em:
This PR expands the "Test and Lint Drivers" chapter in the Lake reference to make `lake test` and `lake lint` substantially easier to follow on a first read. The previous text was a single short paragraph that mentioned `@[test_driver]` once without an example. The new section covers:

* what a test or lint driver actually is — a target (executable, script, or library) that Lake invokes; Lake itself is not a test framework,
* exit-code semantics for executable and script drivers, and the elaboration-error contract for library drivers,
* a verified `lakefile.toml` example for both `testDriver` and `lintDriver`,
* prose instructions for the equivalent `lakefile.lean` configuration (without a code example, since the surrounding chapter is currently in a `:::TODO` because importing Lake's attributes crashes the compiler),
* the single-driver constraint and the conflict between the attribute and the configuration field,
* the `<pkg>/<name>` syntax for drivers defined in dependency packages,
* how `testDriverArgs` and `lintDriverArgs` compose with arguments passed after `--` on the command line,
* the restriction that library test drivers do not accept arguments,
* notes on `lake check-test` and `lake check-lint`.

Motivated by a recent Zulip thread asking for better documentation:
https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Using.20.60lake.20test.60.20effectively/near/598083186

🤖 Prepared with Claude Code


david-christiansen:
Sorry for the slow turnaround here!

I went through and added some more tests/examples and did some light prose editing. I'd like to look at it once more with fresh eyes before a merge.

PR #861, opened 2026-06-10
fix: tutorial header permalinks

david-christiansen:
Bumps dependencies to fix two bugs in tutorial header permalink widgets:
1. They were absolute, rather than relative, URLs, so they ignored `<base>`
2. They had doubled components in their HTML IDs


david-christiansen:
This will also need backporting to v4.30.0

PR #862, opened 2026-06-11
feat: description of do-element elaboration

david-christiansen:


Manual/NotationsMacros/DoElab.lean
```diff
@@ -0,0 +1,1043 @@
...
+{optionDocs backward.do.legacy}
+
+When {option}`backward.do.legacy` is `false`, the extensible elaborator is enabled.
+Custom {keywordOf Lean.Parser.Term.do}`do`-element elaborators extend the desugaring described in {ref "do-notation"}[the section on syntax for monads].
+:::
+
+# Elaboration Overview
+
+The {tech}[syntax kind] `doElem` represents individual {tech}[`do`-elements].
+A sequence of these elements, such as can make up the body of a {keywordOf Lean.Parser.Term.do}`do`-block, is represented by the syntax kind {name}`doSeq`.
```
sgraf812: Broken semantics

Manual/NotationsMacros/DoElab.lean
```diff
@@ -0,0 +1,1043 @@
...
+```leanOutput multiDemo
+("neg", "zero", "pos")
+```
+
+This macro can be adapted to be a {keywordOf Lean.Parser.Term.do}`do`-element by placing it in the `doElem` syntax category and replacing each arm of the multi-way {keywordOf Lean.Parser.Term.doIf}`if` with a {name}`doSeq` instead of a {name}`Term`.
+The syntax definition is almost the same; however, the {keywordOf Lean.Parser.Term.doIf}`else` branch is optional:
+```lean
+syntax (name := multiIf)
+  "if " withPosition(
+    (colGe atomic("|" (atomic(ident " : "))? term) " => " doSeq)+
```
sgraf812, line 124: You could use  `doIfProp` here, that would both allow `let p := ...` conditions and make `mkDoIf` a bit simpler. On the other hand, that loses the nicely striking symmetry to term multi-`if` expansion.

Manual/NotationsMacros/DoElab.lean
```diff
@@ -0,0 +1,1043 @@
...
+In these cases, it may be necessary to define an elaborator.
+:::
+
+::::example "Freezing Mutable Variables with a Macro"
+Within a {keywordOf Lean.Parser.Term.do}`do` block, new {keywordOf Lean.Parser.Term.doLet}`let` bindings may not shadow existing {keywordOf Lean.Parser.Term.doLet}`let mut` bindings.
+However, many mutable variables are not modified after they are initialized.
+It might be convenient to indicate this fact by removing their mutability.
+
+
+There is no existing way to replace a mutable variable with an immutable one, so a macro cannot be a simple {keywordOf Lean.Parser.Term.do}`do`-element.
```
sgraf812: > so a macro cannot be a simple {keywordOf Lean.Parser.Term.do}`do`-element

I don't understand this half sentence, but maybe it's a skill issue.

After reading next sentence: Ah, this refers to the *syntax* of the construct, not the fact that it can or can't be implemented as a macro. so maybe


```suggestion
There is no existing way to replace a mutable variable with an immutable one, so the syntax to do this cannot be a simple {keywordOf Lean.Parser.Term.do}`do`-element.
```

However, a `doElem` it still is; the point is that it opens a block in which mutability changes.

Manual/NotationsMacros/DoElab.lean
```diff
@@ -0,0 +1,1043 @@
...
+#eval show Option Unit from do nothing
+```
+```leanOutput doNothing'
+some ()
+```
+:::
+
+Because the elaborator invokes its continuation explicitly, rather than simply returning a value, it can control the context of elaboration.
+In particular, it can use {name}`withReader` to modify the context, and it can invoke the continuation multiple times in order to support control structures with branching.
+To prevent code size explosions, continuations track whether they may be invoked multiple times in {name Lean.Elab.Do.DoElemCont.kind}`DoElemCont.kind`.
```
sgraf812:
```suggestion
To prevent code size explosions, continuations track whether they may be elaborated multiple times in {name Lean.Elab.Do.DoElemCont.kind}`DoElemCont.kind`.
```

I think this is more accurately reflecting that "invoking the continuation" refers to "invoking the code that elaborates the continuation" rather than "invokes the elaborated continuation"

Manual/NotationsMacros/DoElab.lean
```diff
@@ -0,0 +1,1043 @@
...
+    | throwUnsupportedSyntax
+  let proofStx : Term ←
+    if let some tac := tac? then
+      `(by $tac)
+    else
+      `(by first | contradiction | grind)
+  let proof ← elabTermEnsuringType proofStx (mkConst ``False)
+  dec.elabAsSyntacticallyDeadCode
+  let ty ← mkMonadApp (← read).doBlockResultType
+  return (← Meta.mkAppOptM ``False.elim #[some ty, some proof])
```
sgraf812, line 449:
```suggestion
  return (← Meta.mkAppM ``False.elim #[ty, proof])
```

Manual/NotationsMacros/DoElab.lean
```diff
@@ -0,0 +1,1043 @@
...
+def elabOnce : DoElab := fun stx dec => do
+  let `(doElem| once $body) := stx | throwUnsupportedSyntax
+  let dec ← dec.ensureUnit
+  let info ← inferControlInfoElem stx
+  dec.withDuplicableCont info fun dec => do
+    let returnCont ← getReturnCont
+    let exitCont := dec.continueWithUnit
+    enterLoopBody exitCont exitCont returnCont do
+      elabDoSeq body dec
+```
```
sgraf812: Since you use `withDuplicableCont` here, you don't actually need to return a `numRegularExit` greater than 1 in the handler; any use of `withDuplicableCont` on the outside is redundant. Now that I think of it, it is probably also an unnecessary pessimization in the builtin control info handler for `match` and `if`, because those make the cont duplicable exactly if needed.

If you want to be precise, then in this code, you could should call `inferControlInfoElem body`, do the 
```
  let numRegularExits :=
    bodyInfo.numRegularExits
      + (if bodyInfo.breaks then 2 else 0)
      + (if bodyInfo.continues then 2 else 0)
```
dance and pass the updated info to `withDuplicableCont`. Externally in the control info handler, the control info of `once` should `min numRegularExits 1`.

sgraf812 (commented):
Really nice!! Thank you for writing this, and coming up with such great examples :)

sgraf812:
Also Claude suggested that there a few more occurences of "do item" to rename in Syntax.lean

Manual/NotationsMacros/DoElab.lean
david-christiansen: I think I'll keep it at this one - this isn't a real feature intended for maximal use of the library, it's meant to simply illustrate something with a very parallel construction.

Thanks for the pointer though!

Manual/NotationsMacros/DoElab.lean
david-christiansen: Thanks! I solved it another way, though, by being _e v e n   m o r e  e x p l i c i t_.

Manual/NotationsMacros/DoElab.lean
david-christiansen: This change actually broke it, because the type is usually implicit but needs to be explicit here (`mkAppOptM` is still useful because it finds the universe param)

Manual/NotationsMacros/DoElab.lean
david-christiansen: Thanks, this was very helpful. I fixed the example, which is now much more informative as a result.

PR #863, opened 2026-06-12
feat: tighten inductive FFI description

Kha:
Addresses both #13679 and more general compiler freedom


Manual/Language/InductiveTypes.lean
```diff
@@ -423,6 +423,8 @@ tag := "inductive-types-trivial-wrappers"
 
 If an inductive type has exactly one constructor, and that constructor has exactly one run-time relevant parameter, then the inductive type is represented identically to its parameter.
 
+For a public inductive type under the module system, any private information that prevents these conditions from being tested in other modules will disable this optimization even in the current module.
```
david-christiansen: This took me some time to parse out and think through, so let me unpack my understanding to check for correctness:

If a public inductive type is defined in a module and its constructor is private, then this optimization is not applied **even in the module that defines it**. This is because the ABI of a module's types is determined by the module's interface (that is, its public information). If the optimization were applied only in the defining module, then this would lead to an ABI mismatch. Furthermore, the structure of the private constructor could change without affecting the module's public information. Making the entire type private or its constructor public re-enables the optimization because all clients are able to decide this.

Is that a correct understanding? If so, then presumably one could work around it by having a non-exposed public `def` that names the type externally? If not, then why not?

Manual/Language/InductiveTypes.lean
```diff
@@ -459,64 +461,7 @@ tag := "inductive-types-ffi"
...
-* {name}`S.sc16_2`: {C}`lean_ctor_get_uint16(val, sizeof(void*)*3 + 42)`
-* {name}`S.sc8_1`: {C}`lean_ctor_get_uint8(val, sizeof(void*)*3 + 44)`
-* {name}`S.sc8_2`: {C}`lean_ctor_get_uint8(val, sizeof(void*)*3 + 45)`
-
-::::
-
-::: TODO
-Figure out how to test/validate/CI these statements
-:::
+The exact field layout is implementation-defined; constructor objects should only ever be created, and their fields accessed, by `@[export]`ing a corresponding function from Lean.
```
david-christiansen:
```suggestion
The exact field layout is implementation-defined.
Constructor objects should only be created or unpacked by functions defined in Lean code.
These functions can be made available to C via the {attr}`export` attribute.
```

We don't properly document this attribute yet, but I don't think that needs to block this PR. The `@[export]` syntax is not valid due to the missing name; if we want it to hyperlink once it gets documented, the bare form here is better.

Manual/Language/InductiveTypes.lean
Kha: Correct! But a non-exposed type def is not sufficient as judging its run-time relevancy still requires accessing private data

Kha: Which actually is the reason such a non-exposed type def is currently denied if the RHS is not publicly imported. It would be nice to lift both restrictions at once in the future...

Manual/Language/InductiveTypes.lean
Kha: @david-christiansen It looks like it still expects an argument

Manual/Language/InductiveTypes.lean
david-christiansen: OK, I think the updated text captures the discussion points above.

PR #866, opened 2026-06-18
feat: describe Async

david-christiansen:
This PR adds a chapter about the `Async` monad and fixes some minor issues elsewhere.


Manual/IO/Async.lean
```diff
@@ -0,0 +1,1388 @@
...
+{docstring EAsync}
+
+{docstring BaseAsync}
+
+Infinite loops in {name}`EAsync` and {name}`Async` use a special instance of {name}`ForIn` that ensures that they don't consume stack frames.
+They can therefore be used in long-running asynchronous applications such as servers without the stack overflowing.
+
+Each of these monads has a corresponding type of asynchronous tasks that it can coordinate.
+These tasks can be thought of as handles to an in-flight computation.
+Calling {name}`async` on a monadic action creates a task that runs in the current thread until it suspends, and calling {name}`await` on a task results in a monadic action that waits for the task to complete.
```
algebraic-dev, line 74: `async` can run the computation on another thread, as it uses `BaseIO.asTask` internally without the `sync` flag

Manual/IO/Async.lean
```diff
@@ -0,0 +1,1388 @@
...
+      return "no error"
+    catch e => return (toString e))
+  unless r == "fast failure" do throw (IO.userError "raceFailFast")
+```
+
+## Errors in Event Selection
+
+During selection, errors might occur at any stage of {ref "async-select"}[the protocol].
+Errors thrown by a selector during the initial {name Selector.tryFn}`tryFn` loop terminate the selection immediately.
+An error thrown from a {name Selector.registerFn}`registerFn` or {name Selector.unregisterFn}`unregisterFn`, by contrast, can leave selectors that were already registered without a matching call to {name Selector.unregisterFn}`unregisterFn`.
```
algebraic-dev, line 855: Fixed by https://github.com/leanprover/lean4/pull/14253

david-christiansen: I think what I'll do is wait for after the upcoming RC to avoid coordinating the pre- and post-versions of this chapter across the release. Thanks!

david-christiansen:
I fixed the issues and updated it to the RC. Thanks @algebraic-dev!

PR #882, opened 2026-07-02
doc: remove unrelated example for nested inductives

ia0:
I'm not sure where this came from, but I suspect some accidental copy/paste, because the last sentence is incomplete and the inductive is not nested.


robsimmons:
Thanks!

PR #883, opened 2026-07-07
chore: documentation for new Lake options

robsimmons:
Documents new additions to Lake added in github.com/leanprover/lean4/issues/14254

### DIFF
--- a/Manual/BuildTools/Lake.lean
+++ b/Manual/BuildTools/Lake.lean
@@ -584,7 +584,7 @@ The facets available for modules are:
 
 : `c.o.noexport`
 
- The compiled object file, produced from the C file, with Lean symbols exported.
+ The compiled object file, produced from the C file, without Lean symbols exported.
 
 : `bc.o`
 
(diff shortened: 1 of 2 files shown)


Manual/BuildTools/Lake.lean
```diff
@@ -600,6 +602,14 @@ The facets available for modules are:
 
   A compressed archive (produced via `leantar`) of the module's build artifacts. {TODO}[Document `leantar` in the manual as well]
 
+: `linkInfoExport`
+
+  The arguments that need to be passed to the C compiler if the intent is for the resulting object file to export all of Lean's symbols.
```
tydeu:
```suggestion
  A structure containing the linker arguments, static objects, and dynamic libraries needed to link a module and its dependencies. Objects have Lean symbols exported.
```

Manual/BuildTools/Lake.lean
```diff
@@ -600,6 +602,14 @@ The facets available for modules are:
...
+
+: `linkInfoNoExport`
+
+  The arguments that need to be passed to the C compiler if the intent is for the resulting object file to export all of Lean's symbols.
```
tydeu:
```suggestion
  A structure containing the linker arguments, static objects, and dynamic libraries needed to link a module and its dependencies. Objects do not Lean symbols exported.
```

The most important thing here is that `linkInfoNoExport` does NOT have Lean symbols exported.

tydeu (commented):
I have elaborated a bit on what they do. Feel free to edit the descriptions further if you want.

Manual/BuildTools/Lake.lean
robsimmons: right, thanks for catching that cutpaste

PR #889, opened 2026-07-14
chore: merge main to nightly after v4.32.0 release

robsimmons:
The `mvcgen'` described in the Verso release notes is immediately out of date, since `mvcgen'` disappears in nightly and presumably in v4.33.0-rc1.

This is a larger issue we need to address when @david-christiansen is back. Release notes in the manual are special in that we *do not want them to follow the current lean*. Therefore, we have three options:

 - Don't use verso features in release notes, keep them effectively markdown
 - Use some verso features in release notes, but when they stop working, just remove the verso bits to keep the text the same (that's an unsatisfying path IMO, but it's the one I'm implicitly taking in this PR).
 - Adapt the functionality used in VersoBlog to actually keep the code samples from . (This seems like overkill and won't work for things like ``{tactic}`mvcgen'` `` when that tactic goes away, so this would also need to be combined with one of the previous two approaches.)

### DIFF
--- a/Manual/Releases/v4_32_0.lean
+++ b/Manual/Releases/v4_32_0.lean
@@ -151,7 +151,7 @@ Other notable performance work:
 
 ## Monadic Verification: `mvcgen'` and `grind` Improvements
 
-The {tactic}`mvcgen'` and {tactic}`grind` ecosystem continues to mature:
+The `mvcgen'` and {tactic}`grind` ecosystem continues to mature:
 
 - [#13983](https://github.com/leanprover/lean4/pull/13983) adds `mvcgen' until $t`, where `$t` is a conv-style pattern; verification-condition generation stops as soon as the program matches the pattern. For instance, compare the traces in these two examples:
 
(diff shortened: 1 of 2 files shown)


jcreedcmu:
What about a variant of your second bullet, which is "don't remove the verso bits, but rather change them to be a hardcoded link to the relevant section of the versioned historical reference manual". This still has the disadvantage that we're not using verso functionality to help us maintain the consistency of that link, but it seems comparatively benign to me since arguably we don't ever expect that link to change its destination ever again.

robsimmons:
@jcreedcmu 's principle is one I at least potentially agree we can follow — but it doesn't make a difference here, funnily enough, because, `mvcgen'` is _never mentioned in the manual_ outside the release notes — it sure seems like it was an intentionally undocumented feature until it replaced mvcgen. Which does make it questionable that we included it in the highlights at all, but that ship may be sailed.

Garmelon:
Nothing has happened on `main` since the release, so we can just remove it from the highlights and fast-forward the `v4.32.0` tag.

PR #898, opened 2026-07-16
doc: fix incorrect subsingleton criteria

ia0:
This PR fixes 2 aspects of the subsingleton criteria:
- It's not the constructor parameter _type_ which must be a parameter or index. It's the parameter itself. That's demonstrated by `Eq`.
- A constructor parameter doesn't need to be a proof or a parameter/index, it can also be recursive. That's demonstrated by `Acc`. This example has been added to the list of examples, in addition to the criteria fix.

I'm happy to reword the criteria and example. Note that I used the word "argument" to mean both "parameter" and "index" as is done previously in the reference manual:
https://github.com/leanprover/reference-manual/blob/c8193e06c01c3aa6df516ad2df466cd7573f7f52/Manual/Language/InductiveTypes.lean#L87

Here is a more formal description of the criteria:
https://github.com/digama0/lean-type-theory/blob/0ba178704380d2fad751f020d461b293f47e36d5/axioms.tex#L148

subsingleton


ia0:
Those criterion are actually under discussion and may change, in particular regarding `Acc`: https://leanprover.zulipchat.com/#narrow/channel/621470-lean4lean/topic/Lean.20Kernel.20Renovation/with/620371318

PR #899, opened 2026-07-16
doc: mention left and right to prove disjunction

ia0:


Manual/BasicProps.lean
```diff
@@ -112,7 +112,7 @@ Because {lean}`Sum` is a type, it is possible to check _which_ constructor was u
 {lean}`Or`, on the other hand, forms propositions: terms that prove a disjunction cannot be interrogated to check which disjunct was true.
 In other words, because {lean}`Or` is not a {tech}[subsingleton], its proofs cannot be used as part of a computation.
 
-In a {ref "tactics"}[tactic] proof, disjunctions can be proved using either constructor ({name}`Or.inl` or {name}`Or.inr`) explicitly via {tactic}`apply`.
+In a {ref "tactics"}[tactic] proof, disjunctions can be proved using either constructor ({name}`Or.inl` or {name}`Or.inr`) explicitly via {tactic}`apply`, but {tactic}`left` and {tactic}`right` are more common.
```
david-christiansen:
```suggestion
In a {ref "tactics"}[tactic] proof, disjunctions can be proved using either constructor ({name}`Or.inl` or {name}`Or.inr`) explicitly via {tactic}`apply`.
The {tactic}`left` and {tactic}`right` tactics select the left and right disjuncts.
```

david-christiansen: This is perhaps very nitpicky, but I don't actually have data about what's more common!

Additionally, I think it's good to be a bit redundant about what's a tactic name vs some other kind of name, for readers who have a hard time seeing font differences or browser hovers. So I added "tactic" to the text describing them.

Does this capture what you were going for?

ia0: Yes, I was mostly surprised that `left` and `right` were not mentioned while `constructor` was mentioned for conjunction line 100. Then I just mimicked the sentence which uses "more common" there. Your suggestion is perfectly fine for me.

david-christiansen:
Thanks!

PR #918, opened 2026-08-11
doc: fix typo in floating-point literals example

ia0:
Line 238 is `(Neg.neg (OfScientific.ofScientific 22523 true 4) : Float)` which evaluates to -2.2523, not -2.523.


david-christiansen:
Thank you!

PR #920, opened 2026-08-11
doc: fix typo in example for using local variables in elab_rules

ia0:


david-christiansen:
That's embarassing - thank you for the fix!

PR #933, opened 2026-09-01
doc: add v4.34.0 highlights

juanjomadrigal:
v4.34.0 release notes highlights

### DIFF
--- a/Manual/Releases/v4_34_0.lean
+++ b/Manual/Releases/v4_34_0.lean
@@ -37,11 +37,9 @@ and 22 other changes.
 
 # Highlights
 
-The kernel is the headline of this release: three routes to a proof of `False`, all surfaced by adversarial testing with AI systems, are closed, and a round of defensive checks and resource bounds follows them.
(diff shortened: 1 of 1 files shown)


Manual/Releases/v4_34_0.lean
```diff
@@ -33,6 +35,260 @@ there were 5 refactoring changes,
 2 improvements to the test suite,
 and 22 other changes.
 
+# Highlights
+
+The kernel is the headline of this release: three routes to a proof of `False`, all surfaced by adversarial testing with AI systems, are closed, and a round of defensive checks and resource bounds follows them.
```
nomeata: This paragraph just sounds too much AI written (I hope I'm not wrong here). In particular it sounds like Claude…

May need a human rewrite to sound more natural? Maybe read, delete, and then manually rephrase?

juanjomadrigal: Yes, sounds like it 😅 
I have rewriten it, hopefully sounds more natural now. I'll re-read the other sections to check if there are other parts that sound too artificial, and perhaps make some commit in the following days.
Thank you!

nomeata: Thanks!

